PR #44, opened 2020-07-27
rewrite Compatibility part

yzl210:


yzl210:
First you need to create a class that extends `CompatibilityProvider<Main class of the plugin you want to add support>`,  
  
and write the codes whch add support for the plugin in the class.
  
then use `CompatibilitiesManager.addCompatibility(name of the plugin you want to add support, class)` to add the class to Oraxen.

For example support MythicMobs(already supported):

create a class:
```
import io.lumine.xikage.mythicmobs.MythicMobs;

public class MythicMobsCompatibility extends CompatibilityProvider<MythicMobs>
```
then add the class to Oraxen:
```
CompatibilitiesManager.addCompatibility("MythicMobs", MythicMobsCompatibility.class)
```

PR #47, opened 2020-07-27
General/update reflections

Lauriichan:
Removed every unnecessary reflection and unused method which used reflections as a base.
Added nbt and reflection lib to build.gradle 

I also fixed some warnings, like unused variable c.c


Th0rgal:
Let's go

PR #66, opened 2020-09-01
Fixed black magic

Lauriichan:
Black magic is awesome!


Th0rgal (approved):
white magic is better

PR #76, opened 2020-09-15
This time there is no magic T^T

Lauriichan:
Added the **/oraxen recipe show <all/ingredients/result/name>** command
Improved some stuff c.c


Th0rgal (approved):
magic is good

Issue #81, opened 2020-10-11
JitPack fails bulding Oraxen


DerFrZocker:
I'm trying to integrate Oraxen into one of my plugins, but can't get the dependency, because JitPack cannot find the build artifacts of Oraxen. It seems that JitPack can build Oraxen, but than is unable to find the build jar.
After trying to build it locally on my machine, I get the same results.

I tested it with commit #e11eff0, here is the build log from JitPack: https://jitpack.io/com/github/oraxen/Oraxen/-e11eff09c0-1/build.log

Th0rgal:
Hmm don't you get a .jar when you try to build it on your own computer?

DerFrZocker:
Just tested it a bit more and found my error, when building it locally (forgot to set a task). I have now the jar and will add it manually to my local maven repo. But the error with JitPack still happens, would be nice if you can look into it, since it is very annoying to recompile / updated the jar after a new release.

Th0rgal:
> Just tested it a bit more and found my error, when building it locally (forgot to set a task). I have now the jar and will add it manually to my local maven repo. But the error with JitPack still happens, would be nice if you can look into it, since it is very annoying to recompile / updated the jar after a new release.

for sure, I'll try to understand what is the bug

kristianvld:
I have the same problem. As far as I can see, all jitpack builds have failed. Example log of tag 1.51.1:
https://jitpack.io/com/github/oraxen/Oraxen/1.51.0/build.log

List of all builds:
https://jitpack.io/com/github/oraxen/Oraxen/

My fix also ended up being to use a manually built .jar file, but due to your license, this has made things hard to distribute to other contributors, as they too would have to manually build the dependency for themselves, as distributing a compiled version or any source code is against the license.

Th0rgal:
> I have the same problem. As far as I can see, all jitpack builds have failed. Example log of tag 1.51.1:
> https://jitpack.io/com/github/oraxen/Oraxen/1.51.0/build.log
> 
> List of all builds:
> https://jitpack.io/com/github/oraxen/Oraxen/
> 
> My fix also ended up being to use a manually built .jar file, but due to your license, this has made things hard to distribute to other contributors, as they too would have to manually build the dependency for themselves, as distributing a compiled version or any source code is against the license.

That's right. Any idea what I did wrong with jitpack?

kristianvld:
[...]

PR #84, opened 2020-11-01
Add maven plugin to build.gradle

DerFrZocker:
This change fixes issue #81 by adding the maven plugin to  build.gradle, so that the jar gets installed to the local repository and that jitpack can grab it.


Th0rgal (approved):
thank you

PR #93, opened 2020-12-10
More Update !

DoctaEnkoda:
Update module for support java 15+ (only compile)
Remove duplication WorldGuard
Update latest version module
Add 1.16 because add Mechanics Invisible Item Frame !


Th0rgal (approved):
It looks good to me!

PR #94, opened 2020-12-15
Add better support Metrics and Online Mode

DoctaEnkoda:
Its not a PR with more update, but better stats on bstats for your plugin with online mode because dont support bungeecord online mode.

Add : 
Player has an issue if Oraxen charge first and not protocollib so class is not found


Th0rgal (approved):
lgtm

PR #95, opened 2020-12-16
Important BUG ! 

DoctaEnkoda:
Bad condition with block physics (crash server)

Remove loadbedore patch plugin.yml because broken with other plugins


Th0rgal (approved):
lgtm

PR #124, opened 2021-03-29
[skip ci] Update to Java 16

DoctaEnkoda:
Java 16 has arrived (in final version), and having a compatibility of this version will only be better for people who have updated from Bukkit (because they have updated their dependencies to Java 16).

Does not need a version itself, no need to recompile unless otherwise or users have a problem with Java 16.
I have put Java 8 compatibility even if I am not really a fan of using such an old technology knowing that in a few months Java 17 will be the new LTS.


DoctaEnkoda:
Can not publish on this version

PR #177, opened 2021-06-14
Fix consumable mechanic issue

MrYoup:
Fix consumable mechanic by updating proper item slot when consumable item is in off hand


Th0rgal (approved):
lgtm

PR #186, opened 2021-07-14
Add custom wands

Phoenixvine32908:
I would like to add custom wands guns to this awsome plugin


Th0rgal:
Hey thanks for the pr but it looks huge! Why was it commited by Laura?

PR #201, opened 2021-08-08
Introduce Item updating and fix command arguments

yzl210:
Item update command:
/o update <targets> all/hand

Also fix commands can't select multiple players


Th0rgal (approved):
lgtm

PR #202, opened 2021-08-10
Backpack Mechanic

yzl210:
```
backpack:
  rows: 1-6
  open_sound(optional):
    sound: see Sound.class
    volume(optional): default 1.0
    pitch(optional): default 1.0
```


Th0rgal (changes requested):
Oraxen is trying to rely on the Spigot and ProtocolLib API only. Could you use the bukkit PersistentDataHolder instead of this nbt-api?

Th0rgal:
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/persistence/PersistentDataHolder.html

You can use it to store data inside itemstacks, blocks (thanks to a shaded lib) and entities.

PR #205, opened 2021-08-13
French translation added (Rework)

clementinise:


Th0rgal (changes requested):
It's not the right version: you are not using "

clementinise:
Should be good now 👍

PR #206, opened 2021-08-14
German translation

ArMiN231:


Th0rgal (changes requested):
Great job! Could you add the new log message added in the last commits?

ArMiN231:
> Great job! Could you add the new log message added in the last commits?

Sure thing. Will do that. Just message me on discord so I won’t forget to do that when I’m able to do.

PR #220, opened 2021-09-10
Fix RepairMechanic setting wrong real durability on Damageable items

YouHaveTrouble:
The division often resulted in value in range 0 to 1 and as it was done on ints, it was rounded down resulting in multiplication by 0 and there fore setting damage on Damageable to 0.


Th0rgal (approved):
lgtm

Issue #219, opened 2021-09-10
Small typo / wrong variable used


aerulion:
Hello there,
I was just browsing through the code and wanted to let you know that I spotted a small mistake, I'm pretty sure the blockBreak is meant to be sent to each individual "viewer" instead of multiple times to the same "player" variable:
https://github.com/oraxen/oraxen/blob/6dc2ba278e6b37a60c4bbdee9a81bc5b984cfa5b/src/main/java/io/th0rgal/oraxen/utils/breaker/BreakerSystem.java#L124

Kind regards,
~aerulion

Th0rgal:
Hey there!

Thanks for spotting the bug. I would love to see a pull request so you would join the contributors.

Thomas

PR #224, opened 2021-09-13
fix sending block break multiple times to same player

aerulion:
Fixed bug described in #219

~aerulion


Th0rgal (approved):
lgtm

Th0rgal:
thanks for the pr

PR #250, opened 2021-11-07
add NBTAPI to plugin.yml

zax71:
Add NBTAPI to plugin.yml to stop this error:

`[io.th0rgal.oraxen.OraxenPlugin] [Oraxen] Oraxen | Loaded class de.tr7zw.nbtapi.NBTContainer from NBTAPI v2.8.0 which is not a depend, softdepend or loadbefore of this plugin.`


DoctaEnkoda:
Its useless

PR #251, opened 2021-11-07
Added support for high resolution armor

CraftCoderr:
This pull request adds support for armor texture with higher resolution than default one (64x32). 
Changes is:
- Shader sources added to repository (removed from .gitignore)
- In main shader file texture resolution replaced with placeholder
- In settings.yml added parameter armour_resolution which should correspond to highest resolution texture of armor
- All other armor textures upscaled to configured resolution, if needed
- When resource pack generating, placeholder of texture resolution replaced with actual value from settings


Th0rgal (approved):
it looks great

Th0rgal:
Thanks for the pull request, I appreciate it!

PR #269, opened 2021-12-14
Fixing Maven not being able to find jar

Mowstyl:
If you don't set the classifier to null, "-all" will be added at the end of the jar's name so Maven't won't recognize it.
This causes that, even if JitPack shows the compilation was successful, Maven won't be able to load the dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,6 +71,7 @@ processResources {
 }
 
 shadowJar {
+    classifier = null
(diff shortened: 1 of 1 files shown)


Th0rgal:
I think changing the classifier is deprecated. Is there not an alternative?

Mowstyl:
I've read it has been deprecated in favor of archiveClassifier. I've changed it, it works.

Th0rgal (approved):
lgtm

Issue #272, opened 2021-12-19
BlockPlaceEvent of furnitures is not cancelled


Kenuux:
**Hey there 👋**

During the development of plugins using Oraxen I had to notice that the BlockPlaceEvent of furnitures is not cancelled if there is no space left. So there is no easy way for developers to find out if the placing of a furniture is failed.

I will open a pull request directly afterwards.

Greetings

PR #273, opened 2021-12-19
Fix BlockPlaceEvent of furnitures is not cancelled (#272)

Kenuux:
**Hey there 👋**

My pull request offers the possibility to find out when a furniture could not be placed (lack of space).

It is tested and works perfectly.

See here for more information: https://github.com/oraxen/oraxen/issues/272


Th0rgal:
Hello there!

I am not sure this is necessary. Don't you see the Not Enough Space message in oraxen vanilla builds?

Thanks,
Thomas

Kenuux:
Hi @Th0rgal 

I think I have expressed myself incorrectly in the description of this pull request.
Yes, the message that there is no more space comes and works correctly.
The only problem is that the developer can't find out if there is no more space because the BlockPlaceEvent is not canceled and still fired. With my fix, the event is canceled if the Furniture could not be placed.

I hope I have expressed myself better now. Here is the issue: https://github.com/oraxen/oraxen/issues/272

Kind regards

Th0rgal (approved):
That makes sense, lgtm

PR #285, opened 2022-01-30
Fix checking for wrong material

pustinek:
Fixes the error that is being thrown when breaking a note_block  - Casting it as tripwire


Th0rgal (approved):
lgtm

PR #308, opened 2022-04-03
"fix" piston interactions

### LINKED ISSUE #301: Bug with a custom block being pushed by a piston.
When trying to push a custom block with a piston set the texture back to a notebook. seems to be no fix for this issue. 

if anyone knows a fix please help otherwise it's an Oraxen bug.

Boy0000:
Closes #301 

Currently noteblocks will update their blockstate if pushed onto specific blocks. Tried cancelling this but couldnt get a reliable way.
Not to mention the block will simply drop a noteblock and loose all other behaviour.
Therefore I just set it to cancel them

For tripwires it would normally drop just a string, now it creates a drop from the tripwire

https://user-images.githubusercontent.com/62521371/161433615-a19436ca-4400-4330-8824-4d8738e97217.mp4


Th0rgal (approved):
lgtm

Th0rgal:
Thank you for your contribution, this will be included in next build.

PR #310, opened 2022-04-15
Revert "fix for when clicking with a furniture item on a furniture block that has click actions"

iGabyTM:
Reverts oraxen/oraxen#306
While testing, @lucian929 found out that the fix isn't complete, and I forgot to put the PR on draft 😒 I apologize about that


Th0rgal (approved):
it looks good to me

PR #314, opened 2022-04-17
emote tabcompletion support

Boy0000:
To my knowledge this is the best we could do.
It is rather annoying that it requires a fake player that cant be removed from the tablist but
Added option to let people specify the playericon it should use to help abit

image

image

Config example:
```yml
heart:
  texture: default/chat/heart
  ascent: 8
  height: 8
  chat:
    tabcomplete: true
    tabIcon: c6307390-acda-48f8-8584-42087ad918f4
    placeholders:
      - "<3"
    permission: "oraxen.emoji.heart"

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/commands/CommandsManager.java
+++ b/src/main/java/io/th0rgal/oraxen/commands/CommandsManager.java
@@ -34,6 +34,7 @@ public class CommandsManager {
                 .withSubcommand((new RecipesCommand()).getRecipesCommand())
                 .withSubcommand((new ReloadCommand()).getReloadCommand())
                 .withSubcommand((new DebugCommand()).getDebugCommand())
+                .withSubcommand((new GlyphCommand()).getGlyphCommand())
                 .executes((sender, args) -> {
                     Message.COMMAND_HELP.send(sender);
                 })
--- a/src/main/java/io/th0rgal/oraxen/commands/GlyphCommand.java
+++ b/src/main/java/io/th0rgal/oraxen/commands/GlyphCommand.java
@@ -1,0 +1,53 @@
+package io.th0rgal.oraxen.commands;
+
+import dev.jorel.commandapi.CommandAPICommand;
+import dev.jorel.commandapi.arguments.ArgumentSuggestions;
+import dev.jorel.commandapi.arguments.TextArgument;
+import io.th0rgal.oraxen.OraxenPlugin;
+import io.th0rgal.oraxen.font.Glyph;
+import org.bukkit.ChatColor;
+import org.bukkit.Material;
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.BookMeta;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.bukkit.inventory.meta.BookMeta.Generation.ORIGINAL;
+
+public class GlyphCommand {
+
+    public CommandAPICommand getGlyphCommand() {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Review by lucian929 — DISMISSED** [review]
Looks good, only issue is in their current state these wont load with the Oraxen reload commands.

**Comment by Boy0000** [comment]
Tried implementing reload, but could not get it to clear the tablist before resending the initial packet.
The tablist-icon implementation rn is kinda meh as you need to know both the texture and signature strings, otherwise Steve/Alex

**Comment by Boy0000** [comment]
@Th0rgal some questions on this.

1. Should I keep it only to emoji.yml or allow it based on adding a config option for it?
2. Should I try and make it so you can get the tabicon of of a UUID or just hardcode in an example?
  * Right now it defaults to Steve and lets you specify a texture + signature which most people wont be able to get
  * Does make it easier to get custom heads there tho that arent tied to players
3. Should i work on adding `/oraxen emotes` or something?
   * Right now it is quite impossible to know what emotes exist if you just dont know
4. Hovermsg that show the :glyphname: aswell

My main project has a custom plugin that essentially does 3 and 4 but it is heavily centered around NMS so i have no clue how easy it is. Shouldnt be too hard tho. 

In general aswell 90% of support questions seem to be related to how to use glyphs for ranks and menus.
I think it mainly stems from either not working fully or being a convoluted process with :glyphname: or %placeholders%
Not sure how, but the formentioned plugin more or less lets you just use :glyphname: everywhere except tab/language-file and perhaps scoreboard. (Tho tab might simply be because TAB is on velocity and not main server for us)

**Comment by Th0rgal** [comment]
Hey thanks for the pr,

1) I think you can keep it as it is
2) Hmm, aren't there tools to generate the signature easily?
3) that's a great idea, and this could open something like a book, may be with a cool texture, we could see with avila
4) great idea

Hmm interesting, idk how we could do that

**Comment by Boy0000** [comment]
https://github.com/MineInAbyss/BondrewdLikesHisEmotes

This is what our own project uses. It does all of the above, including "GIFs" via magic and some WIP actual gifs
mostly tho it is very solid as the `:glyphname:` works everywhere but Tab and lang-files

It is very nmsy but it does have alot of stuff i think wouuld be good for oraxen. Mainly making everything with glyphs easier on the user end
making you able to split stuff into however many fonts you want etc. I would not be able to copy most of this as it is more or less all nms and bigbrain beyond me but yeah. Good featureset i think would be good for oraxen in some way shape or form

**Comment by Boy0000** [comment]
Config Examples:
This will have tabcompletion in chat and also show up in `/oraxen emoji list`
```yml
heart:
  code: 3113
  texture: interface/chat/emojis/heart
  ascent: 9
  height: 12
  is_emoji: true
  chat:
    tabcomplete: true
    placeholders:
    - ':heart:'
    permission: oraxen.emojis
```

This will have tabcompletion in chat and wont show up in `/oraxen emoji list`
```yml
heart:
  code: 3113
  texture: interface/chat/emojis/heart
  ascent: 9
  height: 12
  chat:
    tabcomplete: true
    placeholders:
    - ':heart:'
    permission: oraxen.emojis
```
You can also edit the tablist texture by adding: (By default it is "invisible" so best left alone but option is there)
```yml
chat:
    tabcomplete: true
    tab_icon_texture: "something" # Not recommended to change from default unless you know how
    tab_icon_signature: "something" # Not recommended to change from default unless you know how
    placeholders:
    - ':heart:'
    permission: oraxen.emojis
```


https://user-images.githubusercontent.com/62521371/176175166-8399b210-4aab-410c-a193-beeb37777682.mp4

PR #315, opened 2022-04-18
fixes creation of ghost blocks with custom blocks

### LINKED ISSUE #307: Custom Block and String Destroyable in worldguard region (client side)
Title says it all.

Both of these block mechanic will disappear for client side if destroyed in worldguard region, but at the same time will trigger anticheat if player step on an invisible/destroyed custom block

Boy0000:
Closes #307

https://user-images.githubusercontent.com/62521371/163818704-d9af094b-f363-4ff7-8477-19ccd46bc343.mp4


lucian929 (approved):
👍

PR #316, opened 2022-04-20
More farming mechanics

Boy0000:
Adds a "Farm Block" block. Essentially farmland for custom blocks, but will eventually have more interactions.
Includes mechanic for watering can aswell, tho I plan on merging all this into a Farming mechanic by the end of the PR.

Not sure on how it should split the "isFarmBlock" mechanic as they do kinda only exist for noteblocks, tho they could in theory be furniture aswell. thoughts?
First mechanic i add so it probably is badly formatted so please correct anything.

PR #318, opened 2022-04-25
Various bug fixes

Boy0000:
This PR mainly consists of bug fixes for visual and some bigger bugs.

Fixes issue with crops growing past their final stage, and thus breaking.
This is imported from #316 as that PR wont be fully ready to merge for abit longer.

Also fixes client side visual bug with stringblocks, where the client predicts a blockchange that doesnt happen serverside.
Probably a more optimal way to check this and radius probably does not need to be this big but

https://user-images.githubusercontent.com/62521371/165134130-f447ba46-125a-4c86-81a6-ae7a6cc7636b.mp4

Fixes bug where stringblock float if breaking block below them

https://user-images.githubusercontent.com/62521371/165140000-c34cf56a-9e8d-47dc-bcee-b33aa69751a4.mp4


lucian929 (approved):
Looks good

PR #336, opened 2022-05-05
Play animation on furniture click

lucian929:
Little thing, but I think it's a good way to improve furniture interactions so they look more realistic.

I did this in about 30 seconds so please let me know if there are any issues, so far it's working fine for me


Boy0000 (approved):
lgtm

PR #354, opened 2022-05-17
fix string dropping incorrectly from water

Boy0000:
This does rely on a PaperSpigot event, much like the EntityInsideBlockEvent used to cancel those updates.
Not sure what the limit is for not sticking to Spigot so ill leave this here

https://streamable.com/3015x5

(file too big for showing here sadly)


lucian929 (approved):

lucian929:
Ah didn't realize you wanted @Th0rgal to review this specifically, my bad.

Th0rgal (approved):
that makes sense, it looks good to me

PR #356, opened 2022-05-18
fix placing/breaking next to tripwire updating it

Boy0000:
https://user-images.githubusercontent.com/62521371/169089390-2f299a1b-46e6-4f28-87ad-347d124b4c62.mp4


Boy0000:
https://user-images.githubusercontent.com/62521371/169150761-b1f8d376-577f-495d-a78b-2cf918e8bd4f.mp4

Gonna put this on hold for abit as this seems to be an issue with this
Need to manually correct all blockstates after this, but that is a lot of code to do, so wont be done atm

PR #358, opened 2022-05-19
Update default configs

EspartaProJunior:


Boy0000:
Gonna close this as none of the added features are actually something that is needed

PR #369, opened 2022-06-04
[FIX] Mechanic could be null, causing errors in console

pustinek:


Boy0000 (approved):
though this shouldnt be needed due to the checks with the PDCs above, iguess it doesnt hurt lol
might happen for some reason so

pustinek:
Yeah, we were messing with farming mechanic on our server and some how this error creep up on us. Tried to reproduce it on my local test server without any success. So thank for the merge :smile:

PR #384, opened 2022-06-14
Added getInstance for BlockMechanicFactory

ArubikU:


Boy0000 (changes requested):
You're returning an empty variable. `instance` is never defined. Look at the [NoteBlockMechanicFactory](https://github.com/oraxen/oraxen/blob/master/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/noteblock/NoteBlockMechanicFactory.java#L28-L36)

Boy0000:
you seem to have done this correctly in your API PR so ill close this one

PR #390, opened 2022-06-17
Pack sender related fixes and changes.

rafael5gr2:
Made some changes that fix the problem with the pack being loaded multiple times when a player joins the server.

Made the reload command to be able to un-register the join event if the user decides to disable both the send-pack and send-join-message options.

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/pack/dispatch/AdvancedPackSender.java
+++ b/src/main/java/io/th0rgal/oraxen/pack/dispatch/AdvancedPackSender.java
@@ -14,6 +14,7 @@ import org.bukkit.Bukkit;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
+import org.bukkit.event.HandlerList;
 import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerJoinEvent;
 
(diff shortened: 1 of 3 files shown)


Boy0000:
Whilst I know this happens to some, I cant actually replicate it locally.
As mentioned in [this video](https://youtu.be/iXbAgnzQ6UQ?t=199) from a mojang employee, it seems to be an issue with servers
Which will be fixed in 1.19.1?

rafael5gr2:
Hello there,
After some more testing, one part of the problem is indeed the 1.19 client I used. The other part is that in the method `uploadAsyncAndSendToPlayers` in the class [UploadManager](https://github.com/oraxen/oraxen/blob/master/src/main/java/io/th0rgal/oraxen/pack/upload/UploadManager.java), the plugin constantly registers a new PlayrJoinEvent without un-registering the old one every time someone executes the command `/oraxen reload all` and probably `/oraxen reload pack`. This leads to the problem with the pack loading multiple times(in 1.19), but this is a problem for other versions too because if you set the plugin to send a join message for the pack, it sends it multiple times, and to be more specific 1 + (as many times you pressed the reload command). In the changes I made, I firstly un-register the event, and then I am registering the new one. I also made a change that un-registers the event if the user sets both the send_pack and the send_join_message options to false. Here are some testing videos I made:

- [1.18.2 Latest Paper Server + Latest Oraxen release form spigot + ProtocolLib v4.8.0 + 1.18.2 Vanilla Client](https://www.youtube.com/watch?v=Zzfr0TMJoJ0)
- [1.18.2 Latest Paper Server + Latest Oraxen release form spigot + ProtocolLib v4.8.0 + ViaVersion 4.3.1 + 1.19 Vanilla Client](https://www.youtube.com/watch?v=A4jlM1nv9jA)
- [1.18.2 Latest Paper Server + Patched Oraxen + ProtocolLib v4.8.0 + ViaVersion 4.3.1 + 1.19 Vanilla Client](https://www.youtube.com/watch?v=1TmSCRuByEQ)

Boy0000 (approved):
Ah could replicate it now yeah. Yeah does indeed fix it wherever we can fix it so nice. Thanks for the PR 👍

PR #403, opened 2022-07-02
Custom block place/break/hit/step/fall sound

Boy0000:
Replaces vanilla files and plays custom sounds on different actions.
It is not flawless but its fairly decent


Boy0000:
Example of how to specify new sound options
Old versions might have used `BLOCK.STONE.PLACE` but wiki should be updated to reflect this when merged
```yml
Mechanics:
    noteblock:
      break_sound: block.stone.break
      place_sound: block.wool.place
      hit_sound: my.custom.sound
      step_sound: custom.sound.step
      fall_sound: block.wood.fall

Th0rgal (approved):
lgtm

PR #404, opened 2022-07-08
add check for schedulers

Boy0000:
Idea is that they wont run unless theres an item configured with it
Could simply add more options in mechanics.yml but this feels abit more dynamic and more supporting of backwards compat

not the prettiest but easiest way i could find to loop through itemconfigs and look for mechanic X


Th0rgal (approved):
great idea, lgtm

PR #405, opened 2022-07-09
Fixes for the furniture, string and noteblock mechanics.

rafael5gr2:
This PR fixes an error that could happen if someone was in creative mode and was moving a barrier block in their inventory while targeting a furniture block from 6+ blocks afar(This is also present to the string mechanic and noteblock mechanic). [furniture mechanic error](https://youtu.be/HRG7y83hK3s), [string mechanic error](https://youtu.be/hUA4caWPhE4), [noteblock mechanic error](https://youtu.be/I9MtJmJWp9A)

This PR also fixes an issue with the furniture mechanic not clearing the added block data. This could result in the following [problem](https://youtu.be/2P2dMWjeqs8) when you were middle-clicking a barrier block that was added at a location where previously was a furniture block.

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
+++ b/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
@@ -1,5 +1,6 @@
 package io.th0rgal.oraxen;
 
+import com.jeff_media.customblockdata.CustomBlockData;
 import dev.jorel.commandapi.CommandAPI;
 import dev.jorel.commandapi.CommandAPIConfig;
 import io.th0rgal.oraxen.commands.CommandsManager;
@@ -66,6 +67,7 @@ public class OraxenPlugin extends JavaPlugin {
         final PluginManager pluginManager = Bukkit.getPluginManager();
         resourcePack = new ResourcePack(this);
         MechanicsManager.registerNativeMechanics();
+        CustomBlockData.registerListener(this);
         fontManager = new FontManager(configsManager);
         soundManager = new SoundManager(configsManager.getSound());
         OraxenItems.loadItems(configsManager);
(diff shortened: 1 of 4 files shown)


src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/furniture/FurnitureMechanic.java
```diff
@@ -258,7 +258,13 @@ public boolean removeSolid(World world, BlockLocation rootBlockLocation, float o
...
                 }
             }
-            location.getBlock().setType(Material.AIR);
+            final Block block = location.getBlock();
+            final PersistentDataContainer blockData = new CustomBlockData(block, OraxenPlugin.get());
+            blockData.remove(FURNITURE_KEY);
+            blockData.remove(SEAT_KEY);
+            blockData.remove(ROOT_KEY);
+            blockData.remove(ORIENTATION_KEY);
+            block.setType(Material.AIR);
```
Boy0000: This might be an issue with farmblock and stuff aswell.
From talking with the dev of CustomBlockData 2.0.0 was meant to automatically remove this.
Reading over it seems we forgot to register this.

Could you try this and see if farmblock data is also left over?
And also register the listener for it in `OraxenPLugin`, should just be `CustomBlockData.registerListener(OraxenPlugin.get());`.
I think this might handle most of these for us

Boy0000 (changes requested):
raytrace stuff looks good

Th0rgal:
@rafael5gr2 what do you think about Boy suggestion?

rafael5gr2:
Hello,
Sorry for not responding sooner. I was on vacation.
Also sorry for closing and reopening the PR, I messed up with git.
I revoked the fix I implemented that removes the custom block data and added this line of code 
```
CustomBlockData.registerListener(this);
```
in OraxenPlugin as you told me and it seems to fix the issue.

Boy0000:
@rafael5gr2 np
did u get a chance to see if it fixes farmblocks aswell?
they also store stuff in the pdc but im unsure if that would be removed by the CBD listener

rafael5gr2:
@Boy0000 I am not sure if this fixes farmblocks. I haven't tested it.

Boy0000 (approved):

PR #408, opened 2022-07-14
Very useful feature to manage custom fonts

Emibergo02:
This is an ESSENTIAL feature to ease the use of emoji, guis, and custom fonts.
i'll also create the documentation if this pull request will be accepted

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
+++ b/src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
@@ -57,7 +57,6 @@ public class PrintFontCommand {
             return;
         }
         Glyph g = OraxenPlugin.get().getFontManager().getGlyphs().stream().filter(glyph -> glyph.getName().equals(glyphName)).findAny().orElse(null);
-        if (g == null) sender.sendMessage("§cGlyph not found");
         sender.spigot().sendMessage(printClickableMsg(ChatColor.WHITE + g.getName(), String.valueOf(g.getCharacter()), "§r" + g.getCharacter()));
     }
 
--- a/src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
+++ b/src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
@@ -1,0 +1,112 @@
+package io.th0rgal.oraxen.commands;
+
+import dev.jorel.commandapi.CommandAPICommand;
+import dev.jorel.commandapi.arguments.ArgumentSuggestions;
+import dev.jorel.commandapi.arguments.TextArgument;
+import io.th0rgal.oraxen.OraxenPlugin;
+import io.th0rgal.oraxen.font.Glyph;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.chat.ClickEvent;
+import net.md_5.bungee.api.chat.ComponentBuilder;
+import net.md_5.bungee.api.chat.HoverEvent;
+import org.apache.commons.codec.DecoderException;
+import org.apache.commons.codec.binary.Hex;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.command.CommandSender;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+import static net.md_5.bungee.api.chat.HoverEvent.Action.SHOW_TEXT;
+
+public class PrintFontCommand {
+    public CommandAPICommand getPrintFontCommand() {
+        List<String> glyphnames = new ArrayList<>();
+        OraxenPlugin.get().getFontManager().getGlyphs().forEach(glyph -> glyphnames.add(glyph.getName()));
+        return new CommandAPICommand("printfont")
+                .withPermission("oraxen.command.printfont")
+                .withArguments(new TextArgument("glyphname").replaceSuggestions(ArgumentSuggestions.strings(glyphnames.toArray(new String[glyphnames.size()]))))
+                .executes(((commandSender, args) -> {
+                    printHelpTitle(commandSender);
+                    if (OraxenPlugin.get().getFontManager().getGlyphFromName(String.valueOf(args[0])) != null
+                            ||
+                            String.valueOf(args[0]).equals("all")) {
+
+                        printGlyph(commandSender, (String) args[0]);
+
+                    } else printUnicode(commandSender, (String) args[0]);
+                }));
+    }
+
+    private void printGlyph(CommandSender sender, String glyphName) {
+        if (glyphName.equals("all")) {
+            int i = 0;
+            ComponentBuilder cb = new ComponentBuilder();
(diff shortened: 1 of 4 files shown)


Emibergo02:
this is a showcase of the feature.
if you hover it displays the unicode char and if you click it suggest you the char as a command. 
ctrl+a and ctrl+c and you have the ui symbol in the clipboard

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    private void printHelpTitle(CommandSender sender) {
+        sender.sendMessage(ChatColor.RED + "The following text is " + ChatColor.BOLD + ChatColor.DARK_RED + "CLICKABLE and HOVERABLE" + ChatColor.RESET + ChatColor.RED + "!");
+    }
+
+    private BaseComponent[] printClickableMsg(String text, String unicodeChar, String hoverText) {
+        BaseComponent[] components = new ComponentBuilder(text)
+                .event(new ClickEvent(ClickEvent.Action.SUGGEST_COMMAND, unicodeChar))
```
Boy0000: Seems better to simply use COPY_TO_CLIPBOARD action here and then add a hovermsg or helpmessage to reflect that

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+                utf++;
+            }
+            if (componentBuilder.getParts().size() != 0)
+                sender.spigot().sendMessage(componentBuilder.create());
+
+        } catch (DecoderException e) {
+            e.printStackTrace();
+        }
+    }
```
Boy0000: not really sure what the purpose of this is? i assume this PR is meant to skip the point of adding the code into the  website or make it easier to get for people who dont set the code themselves. Personally think this can just be removed in favor of the above

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                    cb = new ComponentBuilder();
+                }
+                i++;
+            }
+            if (cb.getParts().size() != 0)
+                sender.spigot().sendMessage(cb.create());
+            return;
+        }
+        Glyph g = OraxenPlugin.get().getFontManager().getGlyphs().stream().filter(glyph -> glyph.getName().equals(glyphName)).findAny().orElse(null);
+        sender.spigot().sendMessage(printClickableMsg(ChatColor.WHITE + g.getName(), String.valueOf(g.getCharacter()), "§r" + g.getCharacter()));
```
Boy0000: probably better to use ChatColor.RESET instead of §r

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                    } else printUnicode(commandSender, (String) args[0]);
+                }));
+    }
+
+    private void printGlyph(CommandSender sender, String glyphName) {
+        if (glyphName.equals("all")) {
+            int i = 0;
+            ComponentBuilder cb = new ComponentBuilder();
+            for (Glyph glyph : OraxenPlugin.get().getFontManager().getGlyphs()) {
+                cb.append(printClickableMsg("§r[" + ChatColor.GREEN + glyph.getName() + "§r] ", String.valueOf(glyph.getCharacter()), String.valueOf(glyph.getCharacter())));
```
Boy0000: same here, ChatColor.RESET instead of §r

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static net.md_5.bungee.api.chat.HoverEvent.Action.SHOW_TEXT;
+
+public class PrintFontCommand {
+    public CommandAPICommand getPrintFontCommand() {
+        List<String> glyphnames = new ArrayList<>();
+        OraxenPlugin.get().getFontManager().getGlyphs().forEach(glyph -> glyphnames.add(glyph.getName()));
+        return new CommandAPICommand("printfont")
```
Boy0000: think perhaps print_glyphs makes more sense than font here

Boy0000 (changes requested):
minor stuff otherwise lgtm

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
Emibergo02: it is a nightmare searching on the website the correct codes to use.
btw i suggest to standardize the usage of unicodes favouring the [private use area ](https://en.wikipedia.org/wiki/Private_Use_Areas) range

Boy0000: I mean it isnt really bad. You just grab the code you define (if you did so) and copy the unicode.
But I agree the process should be simpler. Just dont see any use for also sending the formatted unicode right here.

If you give players the glyphid, and a click will have them copy the unicode to clipboard is all

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
Emibergo02: the hover is for displaying unicode chars.
since the char may be very large, writing anything in hover text may be hidden by the image

Boy0000: You can just change the `This message is CLICKABLE` message then.
Make it say "Click one of the below to copy the unicode" or something

src/main/java/io/th0rgal/oraxen/commands/PrintFontCommand.java
Emibergo02: do you mean the book feature? the problem is that the book is small and can only contain emojis and not guis because they are too large. and clickable text on books doesn't work on every client (for security reasons)

Boy0000: just mean remove the bottom msg there. Doesnt seem too useful imo when u can click just the glyphID to copy it directly and also hover to see it.

Think the codeportion above is for  that?

image

Emibergo02: the last part "automates" the unicode website part in the creation of new glyphs

Boy0000: Well if u insist it should remain im fine with it, just was confused on its purpose.
Since this would only really be useful if you didnt know your glyphcode as oraxen specified it for you.
but eh nitpick so

Boy0000:
@Emibergo02 considered this accepted, if you wanted to handle the documentation for this aswell that would be awesome 👍 
otherwise ill sort it in a few days

Boy0000 (approved):

PR #416, opened 2022-07-17
Farming change [work in progress]

ClickNinYT:
note: still WIP, some stuff unfinished

This PR add the ability to use bone meal on custom crop and changing the growing delay system a little bit.

First is the ability to use bone meal. 2 new properties for custom crop has been added: `allow_bone_meal` and `bone_meal_chance`. To allow bone meal usable on the crop, set `allow_bone_meal` to true and set the `bone_meal_chance` the chance for bone meal to work (1 to 9, if specify more than 10 the crop will automatically grow) currently this bone meal chance is still WIP.

Second is the growing delay system. Now instead of providing the `delay` tick and `probability`, a new properties has been added to replace that: `grow_chance`. You can specify a int or double and this will be the chance for the crop to grow after the specified tick interval (define in mechanic.yml).

Light boost is updated so it can now actually toggled off (before no matter what you set it will always default to true).

A new properties has been added called `rain_boost`, which increase the grow chance when there is current rain and the crop is exposed to rain.

Both light boost and rain boost give 0.5 additional grow chance each,


ClickNinYT:
Closed this PR because I think it would be better to be a different fork and I yoinked some code from IACroper (the chance calculation) so yeah.

PR #417, opened 2022-07-20
Update mcv to 1.19 & Fixed server crash

TheDeadSnake:
Updated mcv 1.19:
 > Only updated the version in build.gradle

Fixed server crash:
 > In Paper 1.19 build 62 - 66 (possibly more) the Server crashes when e.g. Location.getChunk() or similar Chunk related methods are called if the given Chunk does not yet exist.
 -> Added isLoaded() method to BlockHelper
 -> Added isLoaded() call to all GenericGameEvent methods that seem to be able to crash the Server (Provide not existent Chunk data).
 Side-note: All 'missing' Chunks seem to be in the nether.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,8 +59,8 @@ compileJava {
 dependencies {
     def actionsVersion = '1.0.0-SNAPSHOT'
 
-    compileOnly group: 'org.spigotmc', name: 'spigot-api', version: '1.18-R0.1-SNAPSHOT'
-    compileOnly("io.papermc.paper:paper-api:1.18-R0.1-SNAPSHOT") { exclude group: 'net.kyori' }
+    compileOnly group: 'org.spigotmc', name: 'spigot-api', version: '1.19-R0.1-SNAPSHOT'
+    compileOnly("io.papermc.paper:paper-api:1.19-R0.1-SNAPSHOT") { exclude group: 'net.kyori' }
     compileOnly group: 'com.github.dmulloy2', name: 'ProtocolLib', version: '4.8.0'
     compileOnly group: 'com.github.Hazebyte', name: 'CrateReloadedAPI', version: 'd7ae2a14c6'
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -59,8 +59,8 @@ compileJava {
 dependencies {
     def actionsVersion = '1.0.0-SNAPSHOT'
 
-    compileOnly group: 'org.spigotmc', name: 'spigot-api', version: '1.18-R0.1-SNAPSHOT'
-    compileOnly("io.papermc.paper:paper-api:1.18-R0.1-SNAPSHOT") { exclude group: 'net.kyori' }
+    compileOnly group: 'org.spigotmc', name: 'spigot-api', version: '1.19-R0.1-SNAPSHOT'
+    compileOnly("io.papermc.paper:paper-api:1.19-R0.1-SNAPSHOT") { exclude group: 'net.kyori' }
```
Boy0000, line 63: leave this at 1.18 as we still wanna support it

src/main/java/io/th0rgal/oraxen/utils/BlockHelpers.java
```diff
@@ -311,4 +311,17 @@ public static BlockFace getAnvilFacing(BlockFace face) {
...
+    * Instead, get Chunk location and check with World.isChunkLoaded() if the Chunk is loaded.
+    */
+    public static boolean isLoaded(World world, Location loc) {
+        return world.isChunkLoaded(loc.getBlockX() >> 4, loc.getBlockZ() >> 4);
+    }
+
+    public static boolean isLoaded(Location loc) {
+        return loc.getWorld() != null && isLoaded(loc.getWorld(), loc);
+    }
+
```
Boy0000, line 326: could you also not just do location.getChunk().isLoaded() instead?

Boy0000 (approved):

Boy0000:
nvm read your comment now lgtm

PR #426, opened 2022-07-28
add sounds to furniture

Boy0000:
This adds new entries to sound.yml for  `block.stone.x` same as for noteblocks as barriers inherit stone sound.
Also makes it so if a config entry wants to use `block.stone.x` it is repathed to the custom version automatically


Th0rgal (approved):
lgtm

PR #428, opened 2022-07-30
Fixed a syntactical error in the default english.yml language file wh…

VergilPrime:
…ich causes the pack_url not to parse in the "join" section. Code corrected to fall in line with code found here: https://github.com/KyoriPowered/adventure/search?q=open_url


Boy0000 (changes requested):
commit this into #424 if you can as im already editing this there to my knowledge

PR #431, opened 2022-08-02
Very useful command /oraxen iteminfo

Emibergo02:
This command is basically used for retrieving customModelData id and other item information in-game.


lucian929 (approved):
LGTM

PR #437, opened 2022-08-10
Fix 1.19.2 Startup Issue

eyeb0l:
This should fix the startup error that CommandAPI does not support the server version when loading the plugin on 1.19.2.


Boy0000 (approved):
lgtm

Issue #395, opened 2022-06-22
Items get assigned custom_model_data overlapping with pre-assigned custom_model_data


VergilPrime:
### I have checked...

- [X] I am using the latest version of Oraxen
- [X] I am using the latest version of any dependencies
- [X] I have checked if any similar bug reports exist

### Description

 - Some items configured have custom_model_data set to a value
 - Some items configured do not have a custom_model_data
 - These items share a material

What can happen is that the items with configured cmd can overlap with generated cmd.

This is likely because items are loaded (and cmds assigned) as they are read, to fix this you would need to store the cmd-unassigned items in a list and wait for all the cmd-assigned items to be loaded before attempting to assign cmd values and load those items.

### Steps to reproduce

Create a bunch of items that use paper as a material, starting at the custom_model_data of 1. These items may need to be in another file besides gui for this test.

gui items have paper as a material and may load first, assigning 1, 2, and 3 to these items on pack creation.

Check inside the resulting pack.zip at assets\minecraft\models\item\paper.json and see that some items have overlapping custom_model_data values.

### Plugin Version

1.136.0

### ProtocolLib Version

5.0.0 snapshot b568

### Server Version

purpur 1.19 1663

### How breaking is the bug?

Minor Bug - Plugin completely functional, but features have non-working aspects

Emibergo02:
i had the same problem. it seems that oraxen checks one file at a time for duplicates. so if you have duplicates in different files this can happen

Emibergo02:
are you sure that the items in question do not have the same name/oraxenId?

VergilPrime:
Very sure, initial items that had this issue were the default gui items "arrow_next_icon" and my custom "red_heart".

Emibergo02:
i think the only solution is to queue non-generated custommodeldata to the end of the config load and then generate ids starting from highest model data (like /o highest_modeldata)

Emibergo02:
@Boy0000 do you think you can queue custom-model-id generation to the end of item files reading?

Boy0000:
if someone could supply me with config examples that would make me able to reproduce it.
Dont quite get the issue, cant reproduce it and from my minor overlook the code makes it "impossible" aswell but

Emibergo02:
You have 3 files in items directory: bbb.yml ccc.yml zzz.yml with all items with PAPER as material
Step to reproduce:
1- set automatically_set_model_data: true in settings.yml
2- generate all model ids
[...]

PR #449, opened 2022-08-26
add some more parentmodel options

Boy0000:


Emibergo02:
is this related to #395?

Boy0000:
no not currently, might include it but unable to reproduce it.

Emibergo02:
> no not currently, might include it but unable to reproduce it.

https://github.com/oraxen/oraxen/issues/395#issuecomment-1229465350

Th0rgal:
good idea

PR #465, opened 2022-08-31
OraxenFurniturePlaceEvent

VergilPrime:
Implemented OraxenFurniturePlaceEvent. Fires twice, once before and once after the block and itemframe are created with a boolean 'postCreation' to signify which.


Boy0000 (changes requested):
I just think this should fire after its been placed and if it is cancelled, the frame is removed
Adding a boolean for this makes little sense, and if it was needed i feel it would be 2 completely different events.
Preferably this would just be called when `place()` method is called and if cancelled handle it just like all other event cancellations do already

VergilPrime:
> I just think this should fire after its been placed and if it is cancelled, the frame is removed Adding a boolean for this makes little sense, and if it was needed i feel it would be 2 completely different events. Preferably this would just be called when `place()` method is called and if cancelled handle it just like all other event cancellations do already

Fixed!

PR #469, opened 2022-09-01
backpack mechanic

Boy0000:
Currently doesnt save the PDC after setting the key for some reason.
It registers it as "having" it after, but when closed it doesnt
Probably oversight on how pdc is set when itemstack is sent to another function or something


Boy0000:
Superseeded by #474

PR #471, opened 2022-09-02
Generate shader armor automatically

Boy0000:
This generates the files inside `Oraxen/pack/optifine` just fine, but it doesnt include it in the final zip file.
Not entirely sure atm how that works but guessing its to do with VirtualFile

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/config/Settings.java
+++ b/src/main/java/io/th0rgal/oraxen/config/Settings.java
@@ -26,6 +26,7 @@ public enum Settings {
 
     GENERATE("Pack.generation.generate"),
     ARMOR_RESOLUTION("Pack.generation.armor_resolution"),
+    AUTOMATICALLY_GENERATE_SHADER_COMPATIBLE_ARMOR("Pack.generation.automatically_generate_shader_compatible_armor"),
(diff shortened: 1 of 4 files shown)


Th0rgal (changes requested):
Using VirtualFiles would indeed be better. I will give you an example.

Th0rgal:
Users should not have to handle the optifine files generated by Oraxen, instead we can just add an option in config to automatically add them to the final zip.

You can find an example in the generate() method of ResourcePack.java.
```java
            customArmorsTextures = new CustomArmorsTextures((int) Settings.ARMOR_RESOLUTION.getValue());

            ...

            if (customArmorsTextures.hasCustomArmors()) {
                output.add(new VirtualFile("assets/minecraft/textures/models/armor",
                        "leather_layer_1.png",
                        customArmorsTextures.getLayerOne()));
                output.add(new VirtualFile("assets/minecraft/textures/models/armor",
                        "leather_layer_2.png",
                        customArmorsTextures.getLayerTwo()));
            }
```

You basically just have to get an InputStream for the file you want to write and specify the folder and the file name. Oraxen will add it to the zip for you and it will never be written on disk.

PR #474, opened 2022-09-07
Storage and backpack mechanic

Boy0000:


Th0rgal (approved):
it looks very cool, why not using if instead of triumph? :)

Boy0000:
> it looks very cool, why not using if instead of triumph? :)

if? just used thriumph for gui base as I used it for some other stuff and was relatively easy to deal with
Mostly in terms of cancelling actions we dont want and updating STORAGE type when multiple people are manipulating it
Is there another GUI thing we use already? could switch but thriumph one is pretty nice to work with ngl

PR #477, opened 2022-09-09
Bugfixeronis

Boy0000:


Boy0000:
@Th0rgal thoughts on the config changes here?
The welcome sound imo is just kinda annoying. If its a toggle its better as makes it clearer you can actually customize it
Also the other ones for making the auto cmd and stuff default to true, as people cant seem to set it themselves and get confused when its shuffled around between restarts

Th0rgal (dismissed):
it looks good, but:
- why replacing oraxen:welcome by minecraft:welcome?
- automatically_set_model_data default to true is like disabling the option because it will automatically add a model data to all configs and that can't be reverted. It would make more sense to keep it to false imo.
- same with automatically_set_glyph_code

Boy0000:
Alright on the automatic, and switch to mc namespace is because of custom block sounds i had to move the sounds.json into minecraft namespace so it overrides default sounds properly. So welcome is not under oraxen namespace rn. Makes it harder to import sounds.json but these should just be added to sound.yml and ur good

Could make it automated but not sure thats worth it

PR #482, opened 2022-09-15
Allow insecure protocol for Polymath host

Rubenicos:
I tried to run a Polymath host at localhost with Python in pterodactyl but Oraxen tries to connect with https, so i fixed it


Boy0000:
@Rubenicos merge master in again and ill accept it

Boy0000:
nvm seems it doesnt have any conflict so ill merge it

PR #500, opened 2022-09-29
Fixed NPE

TheDeadSnake:


src/main/java/io/th0rgal/oraxen/utils/drops/Drop.java
```diff
@@ -1,135 +1,135 @@
...
+    }
+
+    public boolean canDrop(ItemStack itemInHand) {
+        return isToolEnough(itemInHand) && isTypeEnough(itemInHand);
+    }
+
+    public boolean isTypeEnough(ItemStack itemInHand) {
+        if (hasMinimalType) {
+            String itemType = itemInHand == null ? "" : getItemType(itemInHand);
+            return !itemType.isEmpty() && hierarchy.contains(itemType)
```
Boy0000, line 68: iirc isEmpty wouldnt check for a "", for that you would use isBlank which checks for "" or strings with only whitespaces

Boy0000 (changes requested):

src/main/java/io/th0rgal/oraxen/utils/drops/Drop.java
TheDeadSnake: [String.isEmpty()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#isEmpty()) Returns true iff length() is 0.
"".length() == 0

Boy0000: guess both works here to isBlank checks for whitespaces aswell which wont be an issue here so

Boy0000 (approved):

PR #505, opened 2022-10-01
New API Stuff

ClickNinYT:
The idea of this Oraxen API is that it provide developers a more friendlier and flexible way of doing stuff like get ItemStack from ID, get Glyph unicode, remove blocks, etc.

Currently there are only these but more coming soon, I guess.

TODO:
* Item API (build itemstack, get item id, and sort of item manipulating and item info stuff)
* Block API (add, remove block, get block info, block manipulation, etc)
* Ability to register stuff at runtime
* and more

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/api/OraxenItemsAPI.java
+++ b/src/main/java/io/th0rgal/oraxen/api/OraxenItemsAPI.java
@@ -1,0 +1,32 @@
+package io.th0rgal.oraxen.api;
+
+import io.th0rgal.oraxen.items.OraxenItems;
+import org.bukkit.inventory.ItemStack;
+
+/**
+ * Class for Oraxen custom items
+ */
+public class OraxenItemsAPI {
+    /**
+     * Return Oraxen item ID of the provided {@link ItemStack}.
+     * <br> May return null if the provided {@link ItemStack} is not an Oraxen item.
+     *
+     * @param itemStack {@link ItemStack} to get the ID
+     * @return Possibly-null Oraxen item ID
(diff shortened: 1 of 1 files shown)


src/main/java/io/th0rgal/oraxen/api/OraxenItemsAPI.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Return ItemStack of the provided Oraxen item ID.
+     *
+     * @param itemID Oraxen item ID
+     * @return {@link ItemStack}
+     */
+    public static ItemStack getOraxenItemStack(String itemID) {
+        return OraxenItems.getItemById(itemID).build();
+    }
+}
```
Boy0000, line 32: Not sure what the point is when you have access to OraxenItems
ur just making it have extra steps now

Boy0000 (changes requested):

src/main/java/io/th0rgal/oraxen/api/OraxenItemsAPI.java
fishiatee: my idea is that it can be accessed in a more friendlier way with clear description and like that
also lazy method in a single package make it more accessable ig

Boy0000: All this does tho is add another class with the API tag at the end
It really doesnt do much. Would say for backwards compatibility, add methods to OraxenItems and add similar classes

This is just extra steps with different methodnames

PR #515, opened 2022-10-05
(Mostly) Rewrote the dryout check

### LINKED ISSUE #463: [BUG] Farm mechanic not working
### I have checked...

- [X] I am using the latest version of Oraxen
- [X] I am using the latest version of any dependencies
- [X] I have checked if any similar bug reports exist

### Description

With the example config on the docs the farm isn't working.
The wet farm block doesn't turn into dry.

### Steps to reproduce

Place a farmblock with the example config on the docs

### Plugin Version

latest

### ProtocolLib Version

latest

### Server Version

1.19

### How breaking is the bug?

Non-breaking Bug - Plugin still usable, but certain features unavailable

TheDeadSnake:
This adds a more similiar farmland behaviour as vanilla. Moreover it fixes #463.

Changes:
 \> Added a 'farmblock_moist' pdc tag to the custom farmland blocks.
 \> When a block is checked, it will search for the heighest 'moistLevel' in the area (3x3) and sets the maxValue - 1 unto itself.
  -\> moistLevel: Water == 5 && Wet vanilla Farmland == 1
 ~~\> Added a unit translator for ms to ticks as in the [Farmblock Mechanic Wiki](https://docs.oraxen.com/mechanics/block-mechanic/block-mechanic#oraxen-item-and-pack-configuration) the custom farmland blocks are setup using ms.~~[Reverted]

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
+++ b/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
@@ -23,7 +23,6 @@ import io.th0rgal.oraxen.utils.armorequipevent.ArmorListener;
 import io.th0rgal.oraxen.utils.breaker.BreakerSystem;
 import io.th0rgal.oraxen.utils.inventories.InvManager;
 import io.th0rgal.oraxen.utils.logs.Logs;
-import io.th0rgal.oraxen.utils.storage.StorageMechanic;
 import io.th0rgal.protectionlib.ProtectionLib;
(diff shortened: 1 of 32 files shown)


Boy0000 (changes requested):
Think we use ticks everywhere and the ms in docs is just a typo essentially
Would rather update the docs than shuffle the handling compared to other uses in other mechanics

Outside of that lgtm

Boy0000 (approved):

PR #541, opened 2022-10-28
Allow upload pack using http and include extra folder into resource pack

yuxuanchiadm:
Allow upload pack using http and include extra folder into resource pack


Boy0000:
@yuxuanchiadm Isn't this VirtualPack just the assets folder? Like `Oraxen/pack/assets/namespace/...` not much else you can include so not sure this is needed

yuxuanchiadm:
Sometimes other plugins generate resources too. It will be nice if we can include them directly : )

Boy0000:
I mean other plugins might generate stuff but you would still import this by adding it to `Oraxen/pack`
Unless you mean files are generated and added during runtime via API which might be abit scuffy, better to just copy paste but

yuxuanchiadm:
If other plugins generate some assets. I want to automatically include them. Instead copy them manually. It will massively simplify test process

PR #575, opened 2022-11-26
DirectionalType "LANTERN" created

Rexblane:
I just created a simple "lantern" DirectionalType to make only up and down variations


Boy0000:
Not sure on the usecase for this. I would assume this is for using with models, like the vanilla lanterns meaning the block needs to be transparent. Which no oraxen noteblock would support, not against this just not sure on the usecase

oliviathevampire:
Would allow for hanging blocks

Boy0000:
It means u can have a block have a different texture when placed underneath another yes
But lanterns arent blocks, theyre models. I only see a real benefit to this if it could be that or atleast transparency was a thing
Not against it but feels like just a similar thing to "SIDEWAYS Mechanic, different block when placed on blockface EAST"
Log, dropper and furnace serve similar functionality to already existing directional MC vanilla blocks

For this i would atleast think it would mean we had a lamp/lantern mechanic for blocks.
Though RN LightAPI is broken so

Tub-Net:
I agree, this would be amazing

PR #608, opened 2022-12-26
sugar cane block mechanic that i added

bckd00r:
https://www.mediafire.com/file/k4p0z48da90k6t7/sugarblock.rar/file

I wanted to add it myself because I thought that the developer did not have much time, but I had a little difficulty because I did not have full control of the structure, but I would be happy if you review it, it would be very useful to have such a feature on the wiki or spigot page.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ plugins {
 apply plugin: 'maven-publish'
 
 group 'io.th0rgal'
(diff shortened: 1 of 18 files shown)


Boy0000:
this is so wrong on so many levels its not worth commenting on what those things are... essentially you would break the entire plugin

PR #627, opened 2023-01-02
fix function `getPrintGlyphCommand`

Slixe:
Missing a `.collect()` call on the Stream to get all mapped values.

I don't know if you want to keep the value `all` as first or not, but we can also instanciate one List by doing this:
```java
final List<String> glyphnames = glyphnames.addAll(OraxenPlugin.get().getFontManager().getGlyphs().stream().map(Glyph::getName).collect(Collectors.toList()));
glyphnames.add("all");
```


Boy0000:
lgtm, believe the tabcomplete is sorted alphabetcally ingame so the listorder does not matter
thanks 👍

PR #664, opened 2023-01-26
Changing the path of the generated model

GliczDev:
Allows you to change the path of the generated model:
Example:
```yaml
Pack:
  generate_model: true
  model_directory: example/models/block #Model in assets/example/models/block/
  parent_model: block/cube_all
  textures:
  - minecraft:block/texture #Texture in assets/minecraft/textures/block/
```

Currently only supports main model (pulling models, blocking models etc. currently don't work)

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
+++ b/src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
@@ -4,6 +4,7 @@ import io.th0rgal.oraxen.config.Settings;
 import io.th0rgal.oraxen.utils.Utils;
 import org.bukkit.configuration.ConfigurationSection;
 
+import java.util.Arrays;
 import java.util.List;
 
 public class OraxenMeta {
@@ -17,10 +18,12 @@ public class OraxenMeta {
     private String castModel;
     private List<String> layers;
     private String parentModel;
+    private String modelDirectory;
     private boolean generate_model;
     private boolean hasPackInfos = false;
     private boolean excludedFromInventory = false;
     private boolean noUpdate = false;
+    private boolean hasCustomModelDirectory = false;
 
     public void setExcludedFromInventory() {
         this.excludedFromInventory = true;
@@ -51,6 +54,12 @@ public class OraxenMeta {
         // If not specified, check if a model or texture is set
         this.generate_model = configurationSection.getBoolean("generate_model", getModelName().isEmpty());
         this.parentModel = configurationSection.getString("parent_model", "item/generated");
+        String modelDirectory = configurationSection.getString("model_directory");
+        if (modelDirectory != null) {
+            this.modelDirectory = modelDirectory;
+            this.hasCustomModelDirectory = true;
+        } else
+            this.modelDirectory = "minecraft/models";
     }
 
     // this might not be a very good function name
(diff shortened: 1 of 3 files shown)


GliczDev:
I noticed that only main model (the generated one) can use `model_directory`, so i reverted my commit.

Everything should work well

src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
```diff
@@ -51,6 +53,7 @@ public void setPackInfos(ConfigurationSection configurationSection) {
         // If not specified, check if a model or texture is set
         this.generate_model = configurationSection.getBoolean("generate_model", getModelName().isEmpty());
         this.parentModel = configurationSection.getString("parent_model", "item/generated");
+        this.modelDirectory = configurationSection.getString("model_directory", "minecraft/models");
```
Boy0000: `minecraft/models` here is wrong, this should be just empty.
Modelpath = namespace:path and that path does not include the models/textures folder

Boy0000: property should not be named `model_directory`as it is confusing
`generated_model_path` or somehting would be more self explanatory

Boy0000 (changes requested):
There is also an option in settings.yml to have models generate according to their texture  path
I suggest this PR should remove that option to rmeove clutter

src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
GliczDev: So how should i get correct path to generate model? By creating a method that will convert this `namespace:path` to normal file path?

Boy0000: this should just be following the same pattern as model and textures
aka if u want ur model to go into assets/namespace/models/something/mymodel.json you put `namespace:something/mymodel`

Not entirely sure how I feel about this property anyway but most wont care to use it since there isnt much reason to but

GliczDev: What do you think about it now?

Boy0000 (changes requested):
rebase this onto the new branch. moved both your PRs to go into 1.152.0

GliczDev:
> There is also an option in settings.yml to have models generate according to their texture path I suggest this PR should remove that option to rmeove clutter

So i should modify `readModelName` method to something like this?
```java
    private String readModelName(ConfigurationSection configSection, String configString) {
        String modelName = configSection.getString(configString);
        ConfigurationSection parent = configSection.getParent();

        if (modelName == null && configString.equals("model") && parent != null)
            return parent.getName();
        else if (modelName != null && modelName.endsWith(".json"))
            return modelName.substring(0, modelName.length() - 5);
        else return modelName;
    }
```

Boy0000:
> > There is also an option in settings.yml to have models generate according to their texture path I suggest this PR should remove that option to rmeove clutter
> 
> So i should modify `readModelName` method to something like this?
> 
> ```java
>     private String readModelName(ConfigurationSection configSection, String configString) {
>         String modelName = configSection.getString(configString);
>         ConfigurationSection parent = configSection.getParent();
> 
>         if (modelName == null && configString.equals("model") && parent != null)
>             return parent.getName();
>         else if (modelName != null && modelName.endsWith(".json"))
>             return modelName.substring(0, modelName.length() - 5);
>         else return modelName;
>     }
> ```

I can remove the option just fix the rebasing here and coflicts

Boy0000:
seems so, I will test this out soon enough 👍 thank you for the PR

Boy0000 (approved):

PR #670, opened 2023-01-28
Add better texture system

GliczDev:
This PR adds a better texture system with support to the old one (so it's optional).
The advantage of this system is that it allows you to use any parent model template (so if you create custom model and you want it to work as template like `block/cube_all` etc., this system allows you to to that!)

Examples:
```yaml
Pack:
  generate_model: true
  parent_model: block/cube
  textures:
    particle: block/texture_north
    down: block/texture_down
    up: block/texture_up
    north: block/texture_north
    south: block/texture_south
    west: block/texture_west
    east: block/texture_east
```
```yaml
Pack:
  generate_model: true
  parent_model: block/cube_bottom_top
  textures:
    top: block/texture_top
    side: block/texture_side
    bottom: block/texture_bottom
```

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
+++ b/src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
@@ -4,6 +4,7 @@ import io.th0rgal.oraxen.config.Settings;
 import io.th0rgal.oraxen.utils.Utils;
 import org.bukkit.configuration.ConfigurationSection;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -43,19 +44,19 @@ public class OraxenMeta {
         this.fireworkModel = readModelName(configurationSection, "firework_model");
         this.pullingModels = configurationSection.isList("pulling_models")
                 ? configurationSection.getStringList("pulling_models") : null;
-        this.layers = configurationSection.getStringList("textures");
-        ConfigurationSection texturesSection = configurationSection.getConfigurationSection("textures");
-        if (texturesSection != null) {
-            Map<String, String> layersMap = new HashMap<>();
-            texturesSection.getKeys(false).forEach(key -> layersMap.put(key, texturesSection.getString(key)));
-            this.layersMap = layersMap;
+
+        if (configurationSection.isList("textures")) {
+            this.layers = configurationSection.getStringList("textures");
+            List<String> layers = new ArrayList<>();
+            this.layers.forEach(layer -> layers.add(this.layers.indexOf(layer), layer.replace(".png", "")));
+            this.layers = layers;
         }
-        // can't be refactored with for each or stream because it'll throw
-        // ConcurrentModificationException
-        for (int i = 0; i < layers.size(); i++) {
-            String layer = layers.get(i);
-            if (layer.endsWith(".png"))
-                layers.set(i, layer.substring(0, layer.length() - 4));
+        else if (configurationSection.isConfigurationSection("textures")) {
+            ConfigurationSection texturesSection = configurationSection.getConfigurationSection("textures");
+            assert texturesSection != null;
+            Map<String, String> layersMap = new HashMap<>();
+            texturesSection.getKeys(false).forEach(key -> layersMap.put(key, texturesSection.getString(key).replace(".png", "")));
+            this.layersMap = layersMap;
         }
 
         // If not specified, check if a model or texture is set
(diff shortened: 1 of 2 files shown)


magicelk235:
Can I download this?

GliczDev:
> Can I download this?

Yes, go to my profile, find oraxen repository and go into it, select branch `better-texture-system`, download code and compile it using gradle.
But remember, that there may be bugs and this version does not have default assets (I mean these,  that are with bought version)

magicelk235:
I have no idea what I need to do here "compile it using gradle"

GliczDev:
> I have no idea what I need to do here "compile it using gradle"
> 

Use google or something to search how to compile gradle project

src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
```diff
@@ -40,6 +43,12 @@ public void setPackInfos(ConfigurationSection configurationSection) {
         this.pullingModels = configurationSection.isList("pulling_models")
                 ? configurationSection.getStringList("pulling_models") : null;
         this.layers = configurationSection.getStringList("textures");
+        ConfigurationSection texturesSection = configurationSection.getConfigurationSection("textures");
+        if (texturesSection != null) {
+            Map<String, String> layersMap = new HashMap<>();
+            texturesSection.getKeys(false).forEach(key -> layersMap.put(key, texturesSection.getString(key)));
+            this.layersMap = layersMap;
+        }
```
Boy0000: realisticly it would be either or here. This would support both
Think it is better to make an if statement here, something like
```java
if (configurationSection.isStringList("textures")
    this.layers = configurationSection.getStringList("textures");
else this.layers = null;

// same for layersMap
```

src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
```diff
@@ -140,6 +149,14 @@ public List<String> getLayers() {
         return layers;
     }
 
+    public boolean hasLayersMap() {
+        return layersMap != null && !layersMap.isEmpty();
```
Boy0000, line 157: believe isEmpty is both for null and empty?

Boy0000 (changes requested):

src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
GliczDev: It'll throw `NullPointerException` bc layersMap is null and it's trying to invoke method from it
However commons-collections `CollectionUtils.isEmpty()` works like this
https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isEmpty-java.util.Collection-

src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
GliczDev: I think this `else` is unnecessary, bc `null` value of this field stays as it was, so what about something like this:
```java
if (configurationSection.isStringList("textures")
    this.layers = configurationSection.getStringList("textures");
ConfigurationSection texturesSection = configurationSection.getConfigurationSection("textures");
if (texturesSection != null) {
    Map<String, String> layersMap = new HashMap<>();
    texturesSection.getKeys(false).forEach(key -> layersMap.put(key, texturesSection.getString(key)));
    this.layersMap = layersMap;
}
```
or this:
```java
if (configurationSection.isStringList("textures")
    this.layers = configurationSection.getStringList("textures");
else if (configurationSection.getConfigurationSection("textures") != null) {
    ConfigurationSection texturesSection = configurationSection.getConfigurationSection("textures");
    Map<String, String> layersMap = new HashMap<>();
    texturesSection.getKeys(false).forEach(key -> layersMap.put(key, texturesSection.getString(key)));
    this.layersMap = layersMap;
}
```
And also, maybe `ModelGenerator` should have 
```java
} else if (oraxenMeta.hasLayers() {
```
at line 19?
<br>
Edit: `configurationSection.getStringList()` returns null if it's not a list, so this if statement with `configurationSection.isStringList()` is useless, it can't be a list and configuration section at one moment

Boy0000: First example is the same, u either want layers or layerMap but never both
Second example looks better, but the else if should be isConfigurationSection
literally means nothing in the end but yeah
Otherwise looks good, though you could also .stream().map() the getKeys part instead of looping through and adding to a new list

Boy0000: Also getStringList will return an empty list, not null so keep the isStringList check

Boy0000 (approved):

Boy0000:
@GliczDev Gonna finish up the PR myself as I wanted to do some tweaks to the file, but yeah all looks good now 👍 
just gonna test it real quick then ill merge it

PR #714, opened 2023-02-13
Add new glyph tab completion

GliczDev:
Adds new glyph tab completion system using 1.19.1 custom chat completion packet, should work on lower versions using old code.
Also fixes null protocolManager field in FontManager and removes double initialization of CommandManager and FontManager.

obraz


Boy0000:
i do think having the actual unicode is better than the :placeholder:
Atleast adding a setting to toggle between the two

Boy0000:
@GliczDev cannot get this to work on 1.19.3 spigot or paper servers
It just does not tabcomplete anything

PR #717, opened 2023-02-19
Fix issue that stops AtlasGenerator working on linux

repeater64:
:)


Boy0000:
thanks @repeater64 ill merge them in manually as i wanna split the storage stuff into 1.153.0
ill revert the storage commit and merge this PR then ill commit it to 1.153.0 after 👍

repeater64:
Yeah I realised you would prefer those two things to be separate pull requests but I didn't realise I had to make a new branch to stop it going into the same pull request and couldn't then figure out an easy way of undoing what I'd done! Glad you're able to deal with it XD

PR #736, opened 2023-03-10
Add compatibility for 8 bit images for custom armours.

RAVINGAR:
Previous issue existed with custom armours, where if you used a source file with a image depth of 8 bits, the pack would attempt to set the appropriate colour pixel, but due to the incorrect colour depth, the wrong colour would show up and therefore the custom armour would not appear in game.

The fix to this was to create a new BufferedImage that is of 32 bit from the original image. This appears to fix the issue.

Also fixed what I believe was a null pointer issue that seemed to occur whenever I booted the plugin.


Boy0000:
Thanks for the PR @RAVINGAR , tested and works 👍 
For future PR's it is a good idea to not use your Forks master branch, as it easily gets muddied up when you update it later
Merging this into the 1.153.1 Release PR

PR #786, opened 2023-04-07
add itemInHand param to all OraxenPlaceEvents

ssomar1607:
New param for OraxenFurniturePlaceEvent to get the item with which the user placed the furniture.


Boy0000:
Not sure what the point is here? The item will always be the actual furniture item. So to get the item just do OraxenItems.getItemById(event.getMechanic().getItemId()).build()
Or event.getPlayer().getItemInMainHand() @ssomar1607

ssomar1607:
> Not sure what the point is here? The item will always be the actual furniture item. So to get the item just do OraxenItems.getItemById(event.getMechanic().getItemId()).build()

No, the item can have additional nbt tags added by other plugins. And by doing what you purpose we lose these information.

>  event.getPlayer().getItemInMainHand()
I don't think it's a long term solution, imagine you will accept off hand in the future.

Let developers know directly with which item is associated the furniture placed.

PR #798, opened 2023-04-21
Fix Lifeleech mechanic killing npcs and giving player health

abhithedev200:
This pr closes the bug [https://github.com/oraxen/oraxen/issues/796], basically players are able to kill npc's using the Lifeleech mechanic and they are getting the health too. so added a check to see if the damaged entity is an npc or not

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/lifeleech/LifeLeechMechanicListener.java
+++ b/src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/lifeleech/LifeLeechMechanicListener.java
@@ -39,12 +39,11 @@ public class LifeLeechMechanicListener implements Listener {
 
         if (event.getEntity() instanceof LivingEntity damaged) {
             if(damaged.hasMetadata("NPC")){
-                // Its a NPC so just cancel the event and return
-                event.setCancelled(true);
+                // Its a Citizens NPC so just ignore
                 return;
             }
 
-            // If its not an npc then only give the health to damager
+            // If its not a Citizens NPC then only give the health to damager
             damager.setHealth(Math.min(health, maxHealth));
 
             
--- a/src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/lifeleech/LifeLeechMechanicListener.java
+++ b/src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/lifeleech/LifeLeechMechanicListener.java
@@ -37,8 +37,17 @@ public class LifeLeechMechanicListener implements Listener {
         double health = damager.getHealth() + mechanic.getAmount();
         double maxHealth = damager.getAttribute(Attribute.GENERIC_MAX_HEALTH).getValue();
 
-        damager.setHealth(Math.min(health, maxHealth));
         if (event.getEntity() instanceof LivingEntity damaged) {
+            if(damaged.hasMetadata("NPC")){
+                // Its a NPC so just cancel the event and return
+                event.setCancelled(true);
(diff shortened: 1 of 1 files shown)


src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/lifeleech/LifeLeechMechanicListener.java
```diff
@@ -37,8 +37,17 @@ public void onCall(EntityDamageByEntityEvent event) {
         double health = damager.getHealth() + mechanic.getAmount();
         double maxHealth = damager.getAttribute(Attribute.GENERIC_MAX_HEALTH).getValue();
 
-        damager.setHealth(Math.min(health, maxHealth));
         if (event.getEntity() instanceof LivingEntity damaged) {
+            if(damaged.hasMetadata("NPC")){
+                // Its a NPC so just cancel the event and return
+                event.setCancelled(true);
+                return;
+            }
```
Boy0000: this feels like it can be done in a better way
Preferably use Protectionlib.canInteract() method to check
Same goes for most of these mechanics

Boy0000 (changes requested):
.

src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/lifeleech/LifeLeechMechanicListener.java
abhithedev200: but that don't fix incompatibility with citizens i guess? don't know much about ProtectionLib but according to citizens wiki they add a metadata called "NPC" to entity's they spawn
image

Boy0000: would rather use actual api methods over this as NPC might, even if rare, be used elsewhere
plus it isnt actually obvious here what ur checking

tldr make a softdepend on citizens and use api to check it. otherwise ill get around to this whenever i get time to it

abhithedev200: yeah sure, but to explain what I am checking. I am basically checking if the "damaged" entity is a npc or not if its an npc then don't do anything just ignore else do the normal functionality of the Lifeleech mechanic

Boy0000: you are checking if it is a Citizens NPC
checking the metadata will probably fix it but it isnt very descriptive of what you are doing
atleast add comments that it is checking if it is a citizens npc

cancelling the event is also not something you should do here
the listeners job isnt to do this and will probably cause issues with other plugins
simply returning is enough

abhithedev200: alright so, i will do these changes

abhithedev200: did the changes, do you still want me to use the actual api to check or just the metadata check is enough?

abhithedev200: also i think its better to add a method called #isNpc(Entity) in ProtectionLib so we can add support to multiple npc plugins other than citizens

Boy0000:
Closing this due to actually not being able to replicate this

PR #806, opened 2023-04-25
Add more recipe loaders

CrazyDev05:
This PR aims to add every recipe type besides brewing


Boy0000 (changes requested):
Needs to add commands and visual versions of all these loaders

CrazyDev05:
now i only need to add the stonecutter, but i will do that tomorrow

CrazyDev05:
the recipe builders and commands are implemented

Boy0000:
merged into #840 thanks

PR #824, opened 2023-05-06
Adventure API fix

GliczDev:
Moves adventure api (and also some other libraries) from implementation to plugin.yml libraries.
Because of this, fully compiled plugin will weight less, and also adventure components added in paper api will be compatible with oraxen.


Boy0000:
I tried this before but it does not work. There were some issues with it thus it had to be removed.
It would fail alot on first time setup so i would prefer just leave it like it is as filesize isnt an issue atm

PR #825, opened 2023-05-06
Adventure Component fix

GliczDev:
**Won't work if #824 not merged!**

Makes item's name and lore use component(s) to support e.g. translatable components.

Should work only on paper (tested with commands on singleplayer, however not with compiled plugin on server, bc I don't have time rn)
obraz

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -42,8 +42,8 @@ repositories {
 dependencies {
     val actionsVersion = "1.0.0-SNAPSHOT"
 
-    compileOnly("org.spigotmc:spigot-api:1.19.4-R0.1-SNAPSHOT")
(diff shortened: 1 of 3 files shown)


Boy0000:
you cant simply change the logic of itemstacks to be components as i do heavily believe that would break on spigot servers.
Itemnames already support minimessage formatting meaning you should be just fine using multi language stuff.
It will be parced to a string anyways

GliczDev:
> i do heavily believe that would break on spigot servers

it will just not work on spigot servers, I'm checking if `Oraxen.get().isPaperServer` is true

> you should be just fine using multi language stuff.

no, legacy format only supports plain text, colors and decorations

Boy0000:
Ur checking if it is paper but ur fundamentally changing the base API to paper. Thus any time otherwise dealing with display names on items etc it wont work

PR #839, opened 2023-05-25
New Reload Command for compatibility with Custom Recipes plugins

BiallyDEV:
It's adding a new command "/oraxen reload all_except_recipes" for reloading everything except oraxen recipes because "/oraxen reload all" break custom recipes from other plugins untill next server restart.


BiallyDEV:
Videos for showcase with normal reload all and with the new one


https://github.com/oraxen/oraxen/assets/107373572/6c46e534-a27d-47a3-96bb-0d4311a7efbd

https://github.com/oraxen/oraxen/assets/107373572/19fa649b-6887-4b72-8a03-6653be95302a

Boy0000:
the syntax here is awful and i would rather just fix the issue with reload breaking other plugins

PR #883, opened 2023-06-14
Update commandapi to allow 1.20.1

jaapieaapie1:


Boy0000:
will come in #859 thanks

PR #900, opened 2023-06-19
release: 1.158.0

Boy0000:
```yml
- Fix Gesture Shader breaking playerheads
- Add `radius_limitation` to LimitedPlacing mechanic
- Fix for displayname without formatting being italic
- Fix strippable OraxenBlocks always being stripped
- Make glyphs take a `char` to specify a specific unicode easier, over a `code`
  * Old configs will automatically convert
  * Setting `code` in new configs will automatically convert to `char` entry
- Remove PrintGlyph command due to the above change
- Make OraxenItems with TIPPED_ARROW material and `disable_vanilla_interaction` not fire
- Disable equipping OraxenItems with HorseArmor material and `disable_vanilla_interaction` 
- Fix issues with SilkTouch drops
```


Boy0000:
It isnt necessary but it is indeed pointless. If you need the unicode, you should go to the config

PR #951, opened 2023-07-11
Add minimum protocol requirement to send resource pack

Rubenicos:
`min_protocol` config on advanced pack send
ViaVersion is used as optional dependency if it's present (the plugin override player protocol, so ProtocolLib cannot detect by itself)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ repositories {
     maven("https://oss.sonatype.org/content/repositories/snapshots") // Because Spigot depends on Bungeecord ChatComponent-API
     maven("https://jitpack.io") // JitPack
     maven("https://repo.dmulloy2.net/repository/public/") // ProtocolLib
+    maven("https://repo.viaversion.com/") // ViaVersion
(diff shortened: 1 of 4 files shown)


Boy0000:
I'm not sure i see the point of this PR? What is the issue this fixes?

Rubenicos:
> I'm not sure i see the point of this PR? What is the issue this fixes?

There are 2 possible cases:
1. When your server is on an older version (for example 1.19.4), but you only want to send the resource pack to players on a higher version (for example 1.16.5).
2. When your server supports older client versions (for example 1.12 and 1.13), but you only want to send the resource pack to players on a certain version or higher (for example 1.14).

Recently on my network a Prison server was opened using MC 1.19.4 with old versions support, to encourage players to download a more recent version (since 40% are still using old versions) we added skins for pickaxes that only players on 1.16.5 or higher can see.

Boy0000:
Oraxen is not aimed to work with ancient versions of MC.
I don't see much benefit in this being in Oraxen. If you want it I suggest you keep it and use your fork
Better solution is just to not allow players on such old versions to join IMO but
Closing

PR #1000, opened 2023-08-02
Replace ProtocolLib with Netty injection

JNNGL:


Boy0000:
closing this for above reasons

PR #1007, opened 2023-08-07
Fixed memory leak with glyph tab completions

lukalt:
This pull requests fixes a severe memory leak related to glyph tab completion. The list of sent tab completions for each player object is not removed when the player disconnects. This will prevent the finalization of the `Player` object and will ramp up memory consumption up rapidly. 

This results in 100% memory consumption after a couple of hours. 

I've attached a screenshot of the yourkit profiler. You can see that the `currentGlyphCompletions` from the `FontManager` retains 3.3 GiB of memory with strong references. The map contains 6687 player objects as a key which corresponds to the number of joins on the server since startup.

image


Boy0000 (approved):
Thanks, yeah didnt consider that player-object would be different.
In theory dont need to remove onQAuit now as ive swapped map to use UUID instead, but ill keep it just to be sure, thanks 👍

lukalt:
Thank you for the fast merge!

PR #1029, opened 2023-08-20
Change checked protection for Thor Mechanic

Fer-Jg:
Instead of checking the protection for the block the player interacts with, I changed it to check for the block where the lightning is going to be summoned on. With a 50 blocks range, the interacted block tends to be a "safe" block even when pointing into a protected area where the player is not supposed to be summoning anything.


src/main/java/io/th0rgal/oraxen/mechanics/provided/combat/spell/thor/ThorMechanicListener.java
```diff
@@ -36,7 +36,8 @@ public void onCall(PlayerInteractEvent event) {
         if (event.getAction() != Action.RIGHT_CLICK_AIR && event.getAction() != Action.RIGHT_CLICK_BLOCK) return;
         if (event.useItemInHand() == Event.Result.DENY) return;
         if (BlockHelpers.isInteractable(block) && event.useInteractedBlock() == Event.Result.ALLOW) return;
-        if (!ProtectionLib.canUse(player, location)) return;
+        //if (!ProtectionLib.canUse(player, location)) return;
+        if (!ProtectionLib.canUse(player, player.getTargetBlock(null, 50).getLocation())) return;
```
Boy0000, line 40: Remove the comment and make a variable with the targetLock instead of reusing it here and later down

Boy0000 (changes requested):
.

PR #1035, opened 2023-08-21
Folia

Boy0000:


oliviathevampire:
You're not checking if it tries to set into a slot that is higher than 53, or technically it should be 45, so you need to check if the items tries to get into a slot that is higher than 45 and if so, add them to the next page. This is technically why addItem is better as it automatically makes it create a new page if it's over 45 items

Boy0000:
Not an issue as thriumph has pagesoze set to 45 so everything after auto filters into further pages

oliviathevampire:
Nope, I have tested this myself with the same code and the configs from my server, and it tries to set items in slot even over 53, that is part of the auto-placed ones, the ones I manually have set works, but those that are automatically set tries to continue.

Boy0000:
If you mean in the mainGui then sure that might be the case here, but subguis should not. Feel free to open a PR with a quick fix. Otherwise ill do it tomorrow or when i have time

oliviathevampire:
> If you mean in the mainGui then sure that might be the case here, but subguis should not. Feel free to open a PR with a quick fix. Otherwise ill do it tomorrow or when i have time

ye I meant the mainGui, not really sure the best way to do it tho

oliviathevampire:
You need to fix the opposite icons for the buttons :D you have arrow if the item is loaded and the item if it's not :D for the next and previous and also the exit :D

PR #1047, opened 2023-09-06
1.162.0

Boy0000:
```yml
- Add 1.20.2 support
- Add NMS functionality to Oraxen
- Add experimental NMS GlyphHandler
  * Makes glyphs work everywhere (<glyph:glyphid> tags and placeholders)
  * Adds support for translatable-tag aka Multi-Lang support
- Improve ItemUpdater by copying all NBT from other plugins
  * Fixes issue with AdvancedEnchantment enchantments being removed
- Add support for GriefPrevention plugin

- Fix reloading items not correctly apply config changes
- Fix glyphs not working when Force Unicode is enabled
- Fix placing vanilla noteblocks changing instrument on certain blocks
- Fix non-barrier furniture being removed incorrectly
- Fix AdminCommand for placing blocks
- Fix placing Lanterns on custom blocks
- Fix Shulker inventory shuffling when placed on custom blocks
- Cancel Villager-Trade if input is OraxenItem
  * Would still work if the villager is specifically asking for the OraxenItem

- Improve WorldEdit-handler
- Cleanup and improve logging feedback messages


oliviathevampire:
Tested this out, but you're not checking if it tries to set into a slot that is higher than 53, or technically it should be 45, so you need to check if the items tries to get into a slot that is higher than 45 and if so, add them to the next page. This is technically why addItem is better as it automatically makes it create a new page if it's over 45 items

Boy0000:
I have no clue what this comment relates to in this Pull Request

oliviathevampire:
> I have no clue what this comment relates to in this Pull Request

Sorry, meant to send it in the folia PR

oliviathevampire:
How are we supposed to mention glyphs in titles and stuff now when you disabled formatting via api?

Boy0000:
Exact same way u did before...?

oliviathevampire:
> 

Tried it but it did not work when I enabled nms

PR #1049, opened 2023-09-07
Fixes some problems with the updated ItemsView

oliviathevampire:
- [x] Fixed arrow and exit icons being opposite
- [ ] Fix trying to set items in slots over 45


Boy0000:
thanks for the PR. Gonna close this as I've gone ahead and pushed a fix to folia-branch already.
This also seems to just skip items with bad slots. Thanks anyway 👍

PR #1184, opened 2024-01-07
Fixes damage error in 1.19.2

UntouchedOdin0:
I'm unable to find the error log now, but I was having a error as it was unable to find this method while running a 1.19.4 test server `Paper version git-Paper-166 (MC: 1.19.2) (Implementing API version 1.19.2-R0.1-SNAPSHOT) (Git: f528f53)`. 

Here's a fix for the error of which I did have though in case it happens for anyone else.


Boy0000:
As readme says please open pr's to develop branch
Also im not sure what this is meant to do?

If the damageItemstack method throws an error it will do what ur change proposes anyway

PR #1188, opened 2024-01-14
Support Folia 

Euphillya:
https://github.com/oraxen/oraxen/issues/1182

Je tiens à informer que le code n'est pas fusionnable ! Il est juste là pour montrer les différences, et peut être compiler pour l'utiliser sur Folia. 

Je ne fournit pas de Jar, c'est à vous de compiler !

Please note that the code is not mergeable! It's just there to show the differences, and can be compiled for use on Folia. 

I don't supply Jar, it's up to you to compile!


Boy0000:
Thanks for the PR 👍 
Folia is coming in 2.0 update and 2..0 branch should mostly work fully atm on folia
This PR, whilst good, has some flaws
For one I am pretty sure it breaks on Spigot due to the Bukkit#getRegionScheduler etc not existing
2.0 uses FoliaLib for this reason, though a basic Utils class should be fine

I dont plan on adding folia support before 2.0, feel free to reopen a PR with changes to the 2.0 branch 👍

Euphillya:
Yes, it's not compatible with Spigot! That's why I said it couldn't be merged. I just made this PR so as not to really impact Oraxen users who want to use it temporarily on Folia until this V2.0.

PR #1198, opened 2024-01-23
fix: stone step sound will play when walking at the edge of block and stone break sound doesn't play for vanilla blocks

Happy-FZM:
Should fixed:
https://github.com/oraxen/oraxen/issues/1195
https://github.com/oraxen/oraxen/issues/1196


Boy0000:
Whilst this stops the sound from playing it doesnt correctly play the sound it should.
Seeing as this is an ancient minecraft bug that was fixed in newer versions, I want to resolve that aswell
Gonna push a commit to this shortly that should resolve both 👍

Boy0000:
Rebased onto develop-branch to fix commit-log
For future PRs please refer to README and branch off your forks master branch 👍 
Thank you for the contribution however! Appreciate it

PR #1201, opened 2024-01-24
New features and improvements

Happy-FZM:
1. Improve isStandingInside checks (Use the correct collision box to determine whether the entity is standing at the target position of the block)
2. Support custom oraxen inventory icon display name

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/items/ItemParser.java
+++ b/core/src/main/java/io/th0rgal/oraxen/items/ItemParser.java
@@ -87,14 +87,14 @@ public class ItemParser {
         return templateItem != null;
     }
 
-    private String parseComponentDisplayName(String miniString) {
+    public static String parseComponentDisplayName(String miniString) {
         if (miniString.isEmpty()) return miniString;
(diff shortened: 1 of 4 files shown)


Boy0000:
thanks for another PR 👍 
went ahead and further tweaked BlokcHelpers#isStandingInside to use BoundingBoxes for more accuracy
Also checking for players being in spectator, as invisible entities themselves should still count against placing there
per vanilla behaviour

Happy-FZM:
> thanks for another PR 👍 went ahead and further tweaked BlokcHelpers#isStandingInside to use BoundingBoxes for more accuracy Also checking for players being in spectator, as invisible entities themselves should still count against placing there per vanilla behaviour

You should also exclude Item entity, so the type should be LivingEntity😄

Happy-FZM:
> True though stuff like ArmorStand and such entities which arent LivingEntity should also count Might just exclude Item then, cant think of anything else

In fact, ArmorStand extends LivingEntity

Boy0000:
Thanks, idk how i forgot that, pushed a commit to fix it ❤️

PR #1203, opened 2024-01-24
fix: minor fix for BlockHelpers#isStandingInside

Happy-FZM:
Exclude some unnecessary entity types and use LivingEntity#isInvisible to exclude spectator.


Happy-FZM:
resolve conflicts

> already done in develop branch also im pretty sure isInvisible isnt just spectator, but counts entities with invisible potion aswell Disagree that this should exclude said



> already done in develop branch also im pretty sure isInvisible isnt just spectator, but counts entities with invisible potion aswell Disagree that this should exclude said

Oh, that's right.

PR #1213, opened 2024-01-28
fix: change furniture drop location to the correct position

Happy-FZM:
Use the entity location item will drop into strange positions, such as the furniture placed on top will drop into second floor.

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/api/OraxenBlocks.java
+++ b/core/src/main/java/io/th0rgal/oraxen/api/OraxenBlocks.java
@@ -253,7 +253,7 @@ public class OraxenBlocks {
             if (forceDrop || player.getGameMode() != GameMode.CREATIVE)
                 drop = noteBlockBreakEvent.getDrop();
 
-            player.getWorld().sendGameEvent(null, GameEvent.BLOCK_DESTROY, block.getLocation().toVector());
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/th0rgal/oraxen/utils/drops/Drop.java
```diff
@@ -153,7 +151,7 @@ public void spawns(Location location, ItemStack itemInHand) {
 
     public void furnitureSpawns(Entity baseEntity, ItemStack itemInHand) {
         ItemStack baseItem = OraxenItems.getItemById(sourceID).build();
-        Location location = baseEntity.getLocation();
+        Location location = BlockHelpers.toBlockLocation(baseEntity.getLocation());
```
Boy0000: Wondering if BlockHelpers#toBlockCenterLocation is more accurate here
Think block location will sometimes spit it out into walls or something

Boy0000 (changes requested):
.

core/src/main/java/io/th0rgal/oraxen/utils/drops/Drop.java
Happy-FZM: The implementation of toBlockCenterLocation also exists in dropItemNaturally, so we do not need to use toBlockCenterLocation
image

PR #1221, opened 2024-02-01
improve: make item drop at broken block location

Happy-FZM:


Boy0000 (commented):
This would mean items drop from the barrier broken in a multi furniture entity
Outside of that ntohing really changes, and prefer having it drop from the furniture origin for simplicities sake

Unless i am missing something

Happy-FZM:
> This would mean items drop from the barrier broken in a multi furniture entity Outside of that ntohing really changes, and prefer having it drop from the furniture origin for simplicities sake
> 
> Unless i am missing something

Yes, but I think for some larger furnitures let item always drop at front of player would get a better experience.

PR #1259, opened 2024-02-29
Patch 1

INotSleep:
URL in README.md was incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ compileOnly 'io.th0rgal:oraxen:VERSION'
 ```
 </details>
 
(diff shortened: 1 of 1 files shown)


Boy0000:
the readme links to the maven repo to use and clicking it opens said url
it is correct

mitchngwen:
Check again. It's my only dependency that having issues.

Boy0000:
> Check again. It's my only dependency that having issues. Maven specifically

maven is just broken for oraxen independent of the link or not
Gradle works without issues and im unable to find the issue with maven sadly

PR #1262, opened 2024-03-02
Fix nms.NMSHandlers.getHandler() is null error on folia

denene12:
There is a error that occurs because of nms.NMSHandlers.getHandler() is being null and cannot use some crucial commands. This change fixes it but even though still the plugin generates a wrong resource pack after this issue has been solved.


Boy0000:
your change here does nothing
if handler isnt null, return handler, otherwise setup
your code is just the opposite

folia is also not supported and by the time 2.0 comes out itll probably be rewritten again

PR #1265, opened 2024-03-03
Check location has furniture entity

Happy-FZM:
Fix: 
https://github.com/oraxen/oraxen/issues/1264

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/api/OraxenFurniture.java
+++ b/core/src/main/java/io/th0rgal/oraxen/api/OraxenFurniture.java
@@ -18,6 +18,7 @@ import org.bukkit.entity.*;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.persistence.PersistentDataContainer;
 import org.bukkit.persistence.PersistentDataType;
+import org.bukkit.util.BoundingBox;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/furniture/FurnitureListener.java
```diff
@@ -205,6 +205,7 @@ private FurnitureMechanic getMechanic(ItemStack item, Player player, Block place
         if (FurnitureFactory.getInstance().isNotImplementedIn(itemID) || BlockHelpers.isStandingInside(player, placed)) return null;
         if (!ProtectionLib.canBuild(player, placed.getLocation())) return null;
         if (OraxenFurniture.isFurniture(placed)) return null;
```
Boy0000: Might be better to simply implement this as an additional check in isFurniture
Meant for the longest time for it to do so

Boy0000 (changes requested):
.

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/furniture/FurnitureListener.java
Happy-FZM: done

PR #1268, opened 2024-03-04
Fix for noteblock & mushroom stem hit sound

Bruhhhh10:


Boy0000 (approved):

Boy0000:
🤡 thank you. brackets are my mortal enemy it seems

PR #1273, opened 2024-03-07
fix: missing migrated items due to same model name

skick1234:
Migrated items sharing the same material, file name, and different paths were being overridden.


Boy0000:
Not entirely sure what this aims to fix, could you provide a more concrete example of what this fixes @skick1234 ?

skick1234:
@Boy0000 Example, I have some duplicated files need to migrate.
```
model1:npc1/head
model1:npc2/head
model2:anothermodel/head
```
If these models use the same material, they will be migrate to `migrated_head` key and only the last model is saved. I'm not on my computer now to check my case but I think you will understand this example.

PR #1293, opened 2024-03-15
Sound fix for when placing wooden doors/pressure plates against custom block

minustenchan:
If you place a wooden door/pressure plate against a custom block, it will always play the wood place sound regardless if it can be placed there or not, this pr fixes that.


Boy0000:
isnt `target.canPlace` all you need here? That *should* cover both cases no?

minustenchan:
> isnt `target.canPlace` all you need here? That _should_ cover both cases no?

Doesn't seem like it covers both cases maybe because it does target.setBlockData(newData) above or something, I tested a lot with that already, but this code works completely fine.

Boy0000:
@minustenchan if you could rebase this onto develop

PR #1300, opened 2024-03-18
2.0

Boy0000:


Euphillya:
Suggestion : Creates an NMS folder where all NMS usages will be in it. This avoids having a lot of files as the versions progress.

PR #1301, opened 2024-03-18
Fix Hat mechanics for Armor stands

Timoti11:
- After https://github.com/oraxen/oraxen/pull/1065/commits/562acbbce5f299b3baa81e7912a887747011e4f0 commit players lost the ability to put hats on armor racks.

!!! IS NOT A COMPLETED CODE - JUST A PARTIAL ROLLBACK TO A WORKING STATE.


Timoti11:
I also noticed that when putting the hat on the armor stand, there is no sound as there is in the normal game. It would be a good solution to add a toggle to the config for this sound, and have it always be when interacting with the armor stand and head slot like in the original game.

Boy0000:
closed in favor of proper fix by calling event and checking if cancelled via EventUtils, thanks tho

PR #1311, opened 2024-03-24
fix: fix break custom block will drop when player are creative mode

Happy-FZM:
fix break custom block will drop when player are creative mode


Boy0000 (approved):

Boy0000:
@Happy-FZM if you could rebase this onto develop

PR #1334, opened 2024-04-09
fix: NullPointerException when using worldedit to copy and paste furniture

uScoutzZ:


core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/worldedit/WorldEditHandlers.java
```diff
@@ -129,7 +129,7 @@ private FurnitureMechanic getFurnitureMechanic(@NotNull BaseEntity entity) {
                     bukkitValues = (Map<String, Tag>) tag.getValue().get("BukkitValues").getValue();
                 } catch (Exception ignored) {
                 }
-                if (bukkitValues == null) return null;
+                if (bukkitValues == null || bukkitValues.get("oraxen:furniture") == null) return null;
```
Euphillya: We call bukkitValues.get("oraxen:furniture") twice in succession, creating a variable directly to avoid double recovery.

Euphillya (changes requested):

Euphillya (approved):

PR #1337, opened 2024-04-12
feat: Rework CustomBlock Breaking System

Boy0000:


Boy0000:
Whilst an improvement in theory, this PR has major issues with correctly displaying the damage state for some reason

Mostly working properly:

https://github.com/oraxen/oraxen/assets/62521371/ddd934a8-c6d1-41fb-9392-5ce746796012

When using the Iron pickaxe which has a slower speedModifier, it does increase breakTime but the stage progresses faster
client assumes block is broken before it then resends it:

https://github.com/oraxen/oraxen/assets/62521371/1eee9681-af19-4497-b7c7-348f0d1d911d

With hand it is the worst:

https://github.com/oraxen/oraxen/assets/62521371/c0b9a60b-cbd4-425d-b927-2b38b1a94833

PR #1339, opened 2024-04-13
Check once if it is a Paper (or Fork) or Folia server.

Euphillya:
Currently, every time we check if it's a Paper server, we check if the class exists. The server does not change when it is running, so if it is yes it will remain YES. I also focused on checking Folia. Also I removed the String check, even if it is much faster than a Class check, we do not need to check later. Especially since forks like Purpur lose performance because it checks the class as a result.


Boy0000:
gonna close this as I myself prefer the current impl of it
this using ints is also weird

PR #1341, opened 2024-04-14
1.174.0

Boy0000:
**[New]**
- Added 1.20.5/1.20.6 support

**[1.20.5/1.20.6]**
- Adds new 1.20.5+ component properties [Docs](https://docs.oraxen.com/configuration/items-advanced#id-1.20.5-specific-properties)
  *  `max_stack_size` - Sets the maximum slot-size of an OraxenItem\
  *  `enchantment_glint_override` - Sets an override-state for the enchantment glint\
  *  `fire_resistant` - Sets whether this OraxenItem is immune to fire and lava\
  *  `durability` - Sets the durability of this OraxenItem\
  *  `hide_tooltips` - Hides all tooltips from the given OraxenItem on hover\
  *  `food` - Makes this item consumable with several different properties
- Deprecated Food- Durability & MiscMechanics for the above-mentioned properties

**[Changes]**
- Added Russian-language file (@Timoti11 )
- Added Korean-language file (@Shihyeon )
- Added Japanese-language file (@KYOCUTI )

**[Fixes]**
- Take-command not working properly
- DuplicationHandler error
- Drop-probability not taken into account
- Wrong parent-model when material is (WARPED_FUNGUS/CARROT)_ON_A_STICK
- OraxenInv not always giving items
- OraxenInv slot errors
- Placing CustomBlocks replacing lanterns etc
- CustomArmor warning of TRIMS on all <1.20 servers even when using SHADER


oliviathevampire:
Can you pleaseeeee add a Components thing which works kind of the same as how it works in commands? I would love for it to be possible to use any components, it could be optional and you can use the Oraxen stuff by default but you could replace most of it by using components, for stuff like custom_model_data, name, etc

PR #1342, opened 2024-04-14
Sound fix for when placing block against cb

minustenchan:
Fixes BlockPlaceEvent being called even though block cannot be placeable there, thus playing a required.wood/stone.place sound

should fix pretty much every issue there is that I know of for placing against custom block

Tested in 1.20.4 and 1.19.4

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/noteblock/NoteBlockMechanicListener.java
+++ b/core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/noteblock/NoteBlockMechanicListener.java
@@ -16,6 +16,7 @@ import io.th0rgal.oraxen.utils.VersionUtil;
 import io.th0rgal.oraxen.utils.breaker.BreakerSystem;
 import io.th0rgal.oraxen.utils.breaker.HardnessModifier;
 import io.th0rgal.protectionlib.ProtectionLib;
+import net.minecraft.world.InteractionResult;
 import org.apache.commons.lang3.Range;
 import org.bukkit.*;
 import org.bukkit.block.Block;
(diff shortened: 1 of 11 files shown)


minustenchan:
actually I thought the arm swing animation bug was fixed when placing custom block against replaceable block, but turns out it's not, got zero clue why it just doesn't swing the arm when placing custom block against replaceable

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/noteblock/NoteBlockMechanicListener.java
```diff
@@ -16,6 +16,7 @@
 import io.th0rgal.oraxen.utils.breaker.BreakerSystem;
 import io.th0rgal.oraxen.utils.breaker.HardnessModifier;
 import io.th0rgal.protectionlib.ProtectionLib;
+import net.minecraft.world.InteractionResult;
```
Boy0000, line 19: This introduces an NMS-class to non-nms packages
In theory this should be fine aslong as this class exists in all versions
If there is an API equivalent I would like to see it used instead

Boy0000 (changes requested):
As a sidenote I think this might be better aimed in general at 2.0
2.0 has tons of changes and as this isnt vital, would rather see it implemented there

Easier than rebasing these changes onto the very changed 2.0 codebase

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/noteblock/NoteBlockMechanicListener.java
syveriea: pretty sure it exists from all the way to 1.18 to 1.20.4

minustenchan:
> As a sidenote I think this might be better aimed in general at 2.0 2.0 has tons of changes and as this isnt vital, would rather see it implemented there
> 
> Easier than rebasing these changes onto the very changed 2.0 codebase

will open a PR for 2.0 👍

PR #1347, opened 2024-04-16
fix: remove mining fatigue, cancel task if BlockPlaceEvent is cancelled

minustenchan:
Fixes removing mining fatigue when player fully mined the block
Cancel the task if BlockPlaceEvent is cancelled, before it would keep the timer running

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/utils/breaker/BreakerManager.java
+++ b/core/src/main/java/io/th0rgal/oraxen/utils/breaker/BreakerManager.java
@@ -89,7 +89,7 @@ public class BreakerManager {
                 activeBreakerData.sendBreakProgress();
 
                 for (ActiveBreakerData alterBreakerData : activeBreakerDataMap.values()) {
-                    if (alterBreakerData.breaker.getUniqueId().equals(breakerUUID)) continue;
+                    if (!alterBreakerData.breaker.getUniqueId().equals(breakerUUID)) continue;
                     if (!alterBreakerData.location.equals(activeBreakerData.location)) continue;
 
                     OraxenPlugin.get().breakerManager().stopBlockBreak(alterBreakerData.breaker);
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/th0rgal/oraxen/utils/breaker/BreakerManager.java
```diff
@@ -89,7 +89,7 @@ private BukkitTask createBreakScheduler(double blockBreakTime, UUID breakerUUID)
                 activeBreakerData.sendBreakProgress();
 
                 for (ActiveBreakerData alterBreakerData : activeBreakerDataMap.values()) {
-                    if (alterBreakerData.breaker.getUniqueId().equals(breakerUUID)) continue;
+                    if (!alterBreakerData.breaker.getUniqueId().equals(breakerUUID)) continue;
```
Boy0000: Idea here is that it would cancel all other tasks of other players that might be breaking the block
The stop digging packet might handle this but im not sure it does
Alternatively I think that handling it via OraxenBlockBreakEvent is possible here now that it correctly handles the timings

Boy0000 (changes requested):

core/src/main/java/io/th0rgal/oraxen/utils/breaker/BreakerManager.java
syveriea:
```java
            } else if (EventUtils.callEvent(new BlockBreakEvent(block, player)) && ProtectionLib.canBreak(player, block.getLocation())) {
                PacketHelpers.removeMiningFatigue(player);
                activeBreakerData.resetProgress();
                activeBreakerData.sendBreakProgress();
```

instead of adding a OraxenBlockBreakEvent listener, we can just do it like this no?

Boy0000: might be enough though theres alot of edge-cases im worrying about
in theory all this should be handled by the client as when the block is removed they should "abort" damaging the block and that event should trigger

PR #1360, opened 2024-04-21
fix: tall blocks breaking when attempting to place custom block while inside tall block

minustenchan:
attempting to place custom block while the player is standing inside the tall block would actually break the tall block below and keep the other half of the tall block flying on the air, made a fix for it here


core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/custom_block/CustomBlockHelpers.java
```diff
@@ -110,9 +112,18 @@ else if (!worldHeightRange.contains(blockAbove.getY()))
                 return;
             }
 
+            BlockData blockBelowData = blockBelow.getBlockData();
+            if (oldData instanceof Bisected) {
+                if (blockBelowData instanceof Bisected)
```
Boy0000: merge these into one if-statement

Boy0000 (changes requested):

PR #1389, opened 2024-05-11
fix: more oraxen-inventory slot shenanigans

Emibergo02:


Emibergo02:
i'm just an ultra-fan of maps

core/src/main/java/io/th0rgal/oraxen/utils/inventories/ItemsView.java
```diff
@@ -129,30 +129,41 @@ private record GuiItemSlot(ItemStack itemStack, Integer slot) {
...
+                                        .setDisplayName(displayName)
+                                        .build();
+                            }
+                        })
+                );
 
         // avoid possible bug if isOraxenItems is available but can't be an itemstack
         if (itemStack == null) itemStack = new ItemBuilder(Material.PAPER).setDisplayName(displayName).build();
         int slot = settings.getInt(String.format("oraxen_inventory.menu_layout.%s.slot", Utils.removeExtension(file.getName())), 0) - 1;
+        System.out.println("Slot: " + slot + " for item: " + itemStack.displayName());
```
Boy0000: Remove this, if debugging use Logs.debug, never System.out.println
Debug is also kept to a minimum in releases

core/src/main/java/io/th0rgal/oraxen/utils/inventories/ItemsView.java
```diff
@@ -129,30 +129,41 @@ private record GuiItemSlot(ItemStack itemStack, Integer slot) {
...
-            itemStack = new ItemBuilder(OraxenItems.getItemById(icon).getReferenceClone())
-                    .addItemFlags(ItemFlag.HIDE_ATTRIBUTES)
-                    .setDisplayName(displayName)
-                    .setLore(new ArrayList<>())
-                    .build();
-        } catch (final Exception e) {
-            try {
-                itemStack = new ItemBuilder(Material.getMaterial(icon.toUpperCase()))
+
+        itemStack = icon
```
Boy0000: This can be changed to either use the provided icon or if there is no valid itemid defined as a slot, just use the first in the category, no need for material logic here

Boy0000 (changes requested):

core/src/main/java/io/th0rgal/oraxen/utils/inventories/ItemsView.java
Emibergo02: Ups

PR #1392, opened 2024-05-13
Edits some build script, adds paper remapping attribute to successfully remapping for 1.20.6

toxicity188:
Hi, i rewrite some build script for 1.20.6

- In manifest, you can nominate your NMS option by using paperweight-mappings-namespace.
- Bumps some version.
- Downloads protocollib to default.

Notice: In current paper build, this error will be happened.
```
java.lang.NoSuchFieldError: Class net.minecraft.server.MinecraftServer does not have member field 'net.minecraft.commands.Commands vanillaCommandDispatcher'
```
This is because of CommandAPI's issue.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,9 +7,9 @@ import java.util.*
 plugins {
     id("java")
     //id("com.github.johnrengelman.shadow") version "8.1.1"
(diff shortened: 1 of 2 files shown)


Boy0000:
I believe the CommandAPI issue is related to Papers new Brigadier commit as of build 65
should be fixed in their next update, thanks though

PR #1396, opened 2024-05-18
Finally makes 1.20.6 nms glyph handler.

toxicity188:
Finally... i delete old source and rewrite one.
녹화_2024_05_19_04_20_46_701


Boy0000:
Gonna close this and move away from this packet-serliazer logic
2.0 will include it in some form but likely via packet-listeners
thanks for the PR either way 👍

PR #1424, opened 2024-06-14
OraxenBlockDropLootEvent for the new API

ErikSzabo:
This PR adds a new non-cancellable event named OraxenBlockDropLootEvent.  This event will be fired after the loot has been dropped from the custom block. It contains information about the triggered mechanic, vanilla Block and Player if applicable. Additionally it also has a List<DroppedLoot> which contains every loot that was dropped before with the dropped amounts.

This event will be useful for plugins that are trying to implement something like a Hypixel collections plugin or generally just for saving/monitoring stats. 

I haven't tested if this actually work in game, since I never used custom block mechanics in Oraxen, but I don't see any reason why wouldn't it work.

I also tried not to break existing APIs. Usually just the return types changed from void in the method signatures.

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/OraxenBlockDropLootEvent.java
+++ b/core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/OraxenBlockDropLootEvent.java
@@ -1,0 +1,55 @@
+package io.th0rgal.oraxen.api.events.custom_block;
+
+import io.th0rgal.oraxen.mechanics.provided.gameplay.custom_block.CustomBlockMechanic;
+import io.th0rgal.oraxen.utils.drops.DroppedLoot;
+import org.bukkit.block.Block;
+import org.bukkit.entity.Player;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class OraxenBlockDropLootEvent extends Event {
+    private static final HandlerList HANDLERS = new HandlerList();
+
+    private final CustomBlockMechanic mechanic;
+    private final Player player;
+    private final Block block;
+    private final List<DroppedLoot> loots;
+
+    public OraxenBlockDropLootEvent(@NotNull final CustomBlockMechanic mechanic, @NotNull final Block block, final Player player, @NotNull List<DroppedLoot> loots) {
+        this.mechanic = mechanic;
+        this.block = block;
+        this.player = player;
+        this.loots = loots;
+    }
+
+    @Nullable
+    public Player getPlayer() {
+        return player;
+    }
+
+    public Block getBlock() {
+        return block;
+    }
+
+    public CustomBlockMechanic getMechanic() {
+        return mechanic;
+    }
+
+    public List<DroppedLoot> getLoots() {
+        return loots;
+    }
+
+    @NotNull
+    @Override
+    public HandlerList getHandlers() {
+        return getHandlerList();
+    }
+
+    public static HandlerList getHandlerList() {
+        return HANDLERS;
+    }
+}
--- a/core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/noteblock/OraxenNoteBlockDropLootEvent.java
+++ b/core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/noteblock/OraxenNoteBlockDropLootEvent.java
@@ -1,7 +1,6 @@
 package io.th0rgal.oraxen.api.events.custom_block.noteblock;
 
 import io.th0rgal.oraxen.api.events.custom_block.OraxenCustomBlockDropLootEvent;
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/OraxenBlockDropLootEvent.java
```diff
@@ -0,0 +1,55 @@
...
+import org.bukkit.block.Block;
+import org.bukkit.entity.Player;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class OraxenBlockDropLootEvent extends Event {
```
Boy0000: Rename to OraxenCustomBlockDropLootEvent
Also add sub-events for noteblock & stringblock mechanics

Boy0000 (changes requested):

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/OraxenBlockDropLootEvent.java
ErikSzabo: Done

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/stringblock/OraxenStringBlockDropLootEvent.java
```diff
@@ -0,0 +1,27 @@
...
+import io.th0rgal.oraxen.mechanics.provided.gameplay.custom_block.stringblock.StringBlockMechanic;
+import io.th0rgal.oraxen.utils.drops.DroppedLoot;
+import org.bukkit.block.Block;
+import org.bukkit.entity.Player;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+public class OraxenStringBlockDropLootEvent extends OraxenCustomBlockDropLootEvent {
+    public OraxenStringBlockDropLootEvent(@NotNull CustomBlockMechanic mechanic, @NotNull Block block, Player player, @NotNull List<DroppedLoot> loots) {
```
Boy0000: should take a stringblock mechanic directly

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/noteblock/OraxenNoteBlockDropLootEvent.java
```diff
@@ -0,0 +1,26 @@
...
+import io.th0rgal.oraxen.mechanics.provided.gameplay.custom_block.noteblock.NoteBlockMechanic;
+import io.th0rgal.oraxen.utils.drops.DroppedLoot;
+import org.bukkit.block.Block;
+import org.bukkit.entity.Player;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+public class OraxenNoteBlockDropLootEvent extends OraxenCustomBlockDropLootEvent {
+    public OraxenNoteBlockDropLootEvent(@NotNull CustomBlockMechanic mechanic, @NotNull Block block, Player player, @NotNull List<DroppedLoot> loots) {
```
Boy0000: should take a noteblock mechanic directly

Boy0000 (changes requested):

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/noteblock/OraxenNoteBlockDropLootEvent.java
ErikSzabo: Changed. On that note, OraxenNoteBlockBreakEvent also takes in CustomBlockMechanic.

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/stringblock/OraxenStringBlockDropLootEvent.java
ErikSzabo: Changed. On that note, OraxenStringBlockBreakEvent also takes in CustomBlockMechanic.

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/noteblock/OraxenNoteBlockDropLootEvent.java
Boy0000: Feel free to patch that aswell, probably an oversight from copy/pasting the events originally

core/src/main/java/io/th0rgal/oraxen/api/events/custom_block/stringblock/OraxenStringBlockDropLootEvent.java
Boy0000: Feel free to patch that aswell, probably an oversight from copy/pasting the events originally

Boy0000 (approved):

PR #1435, opened 2024-06-23
Remove rewrite to http from https in packUrl

kevin-gosa:
This removes the automatic change from https to http in the packUrl and instead uses directly the URL sent by Polymath.


Boy0000:
manually merged into #1437 due to conflicts above, thanks for the PR anyway

PR #1456, opened 2024-07-15
Make Oraxen compatible with JDK 17

toxicity188:
화면 캡처 2024-07-15 225139
This commit will force Oraxen to build with class version 61(Java 17) even it's toolchain is 21.  
This will make it compatible with 1.20.1 or lower version of Bukkit.


Boy0000:
oraxen works fine on older versions aslong as the server is using java21
ill consider this and make it work on jdk17 still unless it brings issues like before with using 1.21 API

PR #1468, opened 2024-07-30
fix: thread deadlock issue with breaker system

Speedy11CZ:
Until now, the breaker system packet listener was executed directly in the Netty thread.
At the same time, the breaker system synchronizes with the main thread when accessing the block type. 
If all netty threads are occupied by the breaker system, a race condition may occur, with the main thread waiting for a free netty thread and the breaker system waiting for a free main thread. 
As a result, the entire server will deadlock. With lower TPS, there is a greater chance that a deadlock will occur.

This fix uses an asynchronous packet listener that is supported by the ProtocolLib plugin. This prevents the netty thread from being busy while waiting for the main thread, and thus deadlock race condition cannot occur.

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
+++ b/core/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
@@ -56,6 +56,7 @@ public class OraxenPlugin extends JavaPlugin {
     private InvManager invManager;
     private ResourcePack resourcePack;
     private ClickActionManager clickActionManager;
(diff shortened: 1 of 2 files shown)


Boy0000:
image

lilmayu:
I'm trying to learn netty coding in MC/Bukkit. May I ask why this would "increase ping in rereciving those packets"? To my knowledge, using an ProtocolLib async listener should NOT increase the ping in the contrast of synchronous one (the original implementation), no?

PR #1550, opened 2024-10-22
Add OraxenPackCreateEvent

Euphillya:
This adds an event when the texture pack is ready. This is useful if you use an external resource pack generator that does a merge (like ResourcePackManager)

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/api/events/OraxenPackCreateEvent.java
+++ b/core/src/main/java/io/th0rgal/oraxen/api/events/OraxenPackCreateEvent.java
@@ -1,0 +1,67 @@
+package io.th0rgal.oraxen.api.events;
+
+import io.th0rgal.oraxen.utils.VirtualFile;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+/**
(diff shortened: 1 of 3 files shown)


Boy0000:
this event is pointless as OraxenPackGeneratedEvent is called right before and does what this does?

Euphillya:
> this event is pointless as OraxenPackGeneratedEvent is called right before and does what this does?

No, he is called before.

Boy0000:
Oraxens event is called first, then ur just calling this new event, there is no difference here as i see anything other than the name of the event

image

Boy0000:
there is no reason to do anythign after pack.zip is made
The event oraxen calls now if meant for plugins to tweak the RP before its finished
your event here does nothing important other than say "pack is now zipped" which is not relevant
its a pointless middle-event between OraxenPackGeneratedEvent and OraxenPackUploadedEvent

not gonna merge this one

Euphillya:
> there is no reason to do anythign after pack.zip is made The event oraxen calls now if meant for plugins to tweak the RP before its finished your event here does nothing important other than say "pack is now zipped" which is not relevant its a pointless middle-event between OraxenPackGeneratedEvent and OraxenPackUploadedEvent
> 
> not gonna merge this one

Yeah, well it's useless. I'm sorry to tell you, but since the change of lead developer, I feel like it's getting worse than better.

PR #1649, opened 2025-08-20
Fix Music Disc stuff

Hugo5000:
This one should fix most music disc related stuff.

Like music discs not stopping when playing in a turntable
Turntables with volume of 1 and pitch of 1 now also act like normal jukeboxes, i.e. play the disc like a jukebox.

Tho there is one case where a jukebox will land in limbo if an old itemstack can't be parsed from the pdc.
at least now I catch the exception and return null

If I forgot a music disc related issue, let me know.


Th0rgal (approved):
Thank you Hugo, this lgtm. How much time would you estimate it took you?

PR #1652, opened 2025-08-28
Add compaitibility for PacketEvents

Hugo5000:
This PR adds compatibility for the use of PacketEvents as an alternative to ProtocolLib

the code should do 1:1 what ProtocolLib would do except parsing of snapshot cannot be done reliably since how it was done was to use the field of ProtocolLib to get the release date of the newest version.

I would appreciate if someone tested this before releasing tho.

this is based on #1650 so merge that first.

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/packets/PacketEventsAdapter.java
+++ b/core/src/main/java/io/th0rgal/oraxen/packets/PacketEventsAdapter.java
@@ -1,0 +1,75 @@
+package io.th0rgal.oraxen.packets;
+
+import com.github.retrooper.packetevents.PacketEvents;
+import com.github.retrooper.packetevents.event.PacketListener;
+import com.github.retrooper.packetevents.event.PacketListenerCommon;
+import com.github.retrooper.packetevents.event.PacketListenerPriority;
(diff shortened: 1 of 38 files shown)


core/src/main/java/io/th0rgal/oraxen/packets/PacketEventsAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+    private PacketListenerCommon efficiencyMechanicListener;
+
+    @Override
+    public boolean isEnabled() {
+        return PacketAdapter.isPacketEventsEnabled();
+    }
+
+    @Override
+    public void registerInventoryListener() {
+        inventoryPacketListener = register(new InventoryPacketListener(), PacketListenerPriority.MONITOR);
```
Hugo5000, line 35: should we do a check here just in case it never gets registered twice? well it should never happen bc it only is registered in the onEnable but still

Th0rgal: Since it should never happen i think the best is to check it in order to log a severe error if it happens

PR #1669, opened 2025-10-22
feat: added support for minecraft 1.21.4+ item model definition properties

pxlarified:
Allows more model-definitions to be usable.

Add support for the following item model definition properties:
- oversized_in_gui: Allows items to extend beyond slot boundaries in GUI when scaled up
- hand_animation_on_swap: Controls whether hand animation plays when swapping items (only when switching from an item to another item)
- swap_animation_scale: Adjusts the scale of the swap animation

The main feature is the oversized_in_gui property.
These properties can now be configured in item pack settings and will be properly included in the generated resource pack.

This feature was added after a request from a user on discord.

**Configuration**
```yml
oversized_item:
  itemname: <yellow>Oversized Item
  Pack:
    generate_model: false
    model: item/large_item
    oversized_in_gui: true
```

**Model - Used for the oversized in gui as an example**
```json
{
	"parent": "item/handheld",
	"textures": {
		"layer0": "default/welcome_disk"
	},
	"display": {
		"gui": {
			"translation": [-2, -2, 0],
			"scale": [2, 2, 2]
		}
	}
}
```
image

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
+++ b/core/src/main/java/io/th0rgal/oraxen/items/OraxenMeta.java
@@ -37,6 +37,9 @@ public class OraxenMeta {
     private boolean excludedFromCommands = false;
     private boolean noUpdate = false;
     private boolean disableEnchanting = false;
+    private boolean oversizedInGui = false;
+    private boolean handAnimationOnSwap = true;
+    private float swapAnimationScale = 1.0f;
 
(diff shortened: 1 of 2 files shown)


pxlarified:
Hi @Th0rgal,
Thank you for the offer! You can reach me via Discord at `the_jan_craft`. I'm already part of the Oraxen support team.

Best regards,
Jan

PR #1758, opened 2026-01-27
feat: add Cross-version shader overlay support for 1.21.4+ compatibility

pxlarified:
## 🟠 Cross-version shader overlay support for 1.21.4+ compatibility

- **Summary** - Adds cross-version shader compatibility for Minecraft 1.21.4+ by generating overlay shaders and updating pack.mcmeta format handling
- **Description** - Implements shader overlay system to support different client versions accessing the same resource pack, addressing breaking changes in Minecraft 1.21.4+ shader format requirements

- **Changes**
  - Added `TextShaderTarget` class for version-specific shader generation with pack format constants
  - Enhanced `ResourcePack.java` with overlay shader generation for cross-version compatibility
  - Updated pack.mcmeta generation to include `supported_formats` and overlay entries
  - Simplified font system by removing configurable default font setting
  - Refactored `GlyphAppearance.fromConfig()` to use standard minecraft:default font
  - Added shader overlay directories (`overlay_1_21_4_5`, `overlay_1_21_6_plus`) for different client versions
  - Fixed shader JSON generation to use fully-qualified core references (`minecraft:core/shader_name`) for 1.21.4+

- **Verification**
  - Resource pack generates correctly for both 1.21.4/1.21.5 and 1.21.6+ server versions
  - Overlay shaders are created in appropriate directories when server version differs from client capabilities
  - pack.mcmeta includes proper supported_formats range and overlay entries
  - Font generation maintains backward compatibility while using simplified default font approach
  - Tested on multiple client and server versions.

### DIFF
--- a/core/src/main/java/io/th0rgal/oraxen/pack/generation/ResourcePack.java
+++ b/core/src/main/java/io/th0rgal/oraxen/pack/generation/ResourcePack.java
@@ -79,6 +79,7 @@ public class ResourcePack {
         textShaderFeatures = null;
         textEffectSnippets = null;
         textEffectSnippetsTarget = null;
+        shaderOverlaysGenerated = false;
 
         makeDirsIfNotExists(packFolder, new File(packFolder, "assets"));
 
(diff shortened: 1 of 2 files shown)


pxlarified:
**NOTE**
While this works with 1.21.4 servers and generates shaders compatible with both 1.21.4 and 1.21.5, it does not support 1.21.6 (unless you are on that server version) or later shader formats.

PR #1813, opened 2026-04-24
fix: pack/lang glyph double parsing

pxlarified:
## 🟢 Fix pack/lang glyph double parsing
- **Summary** - Changed pack JSON processing so `pack/lang` string values are parsed structurally instead of running the whole document through MiniMessage.

- **Description** - This removes the path that could expand `<glyph:...>` content during pack generation and then let the same glyph be handled again later, which was causing the glyph to appear twice.

- **Changes** - Updated `core/src/main/java/io/th0rgal/oraxen/pack/generation/PackFileCollector.java` to parse JSON with Gson, recurse through objects and arrays, and only apply `AdventureUtils.parseLegacyThroughMiniMessage(...)` to string values.


Th0rgal:
Superseded by #1815, which merges this branch (preserving every commit) together with the other three open PRs from @pxlarified and adds follow-up fixes addressing the review feedback. Thanks for the contribution — closing in favor of the combined PR.

PR #1814, opened 2026-04-25
fix: Furniture Barrier Cleanup

pxlarified:
## 🟠 Furniture Barrier Cleanup Fix
- **Summary** - Fixed furniture barrier removal so offset barrier blocks are cleaned up when one barrier is broken.
- **Description** - The furniture removal path now uses the same centered root location as placement and also scans nearby marked barrier blocks tied to the same base entity, which prevents leftover barriers from surviving after a single hitbox block is broken.

- **Changes** - Updated `core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/furniture/FurnitureMechanic.java` to normalize the removal root location, extract shared barrier block removal into a helper, and add a fallback scan for marked barrier blocks owned by the same furniture entity.


Th0rgal:
Superseded by #1815, which merges this branch (preserving every commit) together with the other three open PRs from @pxlarified and adds follow-up fixes addressing the review feedback. Thanks for the contribution — closing in favor of the combined PR.

PR #1831, opened 2026-05-17
Oraxen v1.215.0

pxlarified:
### 🟠 fix: Folia Shaped Block Chunk Conversion
- **Summary** - Schedule shaped-block waxed copper chunk conversion on the chunk's region thread instead of the Folia global scheduler.
- **Description** - Newly generated chunks can trigger waxed copper conversion after load. On Folia 1.21.11, running that delayed conversion globally can read block state from the wrong context and hit a null `Level.getCurrentWorldData()` path. The conversion now uses `SchedulerUtil.runAtLocationLater` with the chunk location and skips work if the chunk unloads before the delayed task runs.

- **Changes** - Updated `ShapedBlockMechanicListener#onChunkLoad` to dispatch conversion through the region scheduler, and added an `isLoaded` guard before scanning chunk blocks.

### 🟠 fix: Optional PacketEvents Startup
- **Summary** - Keep Oraxen command registration from hard-requiring PacketEvents when the optional plugin is not installed.
- **Description** - Oraxen 1.214.0 added the totem animation command with direct PacketEvents class references. Servers without PacketEvents could throw `NoClassDefFoundError: com/github/retrooper/packetevents/wrapper/PacketWrapper` while enabling Oraxen, disabling the plugin and preventing players from joining through setups such as Velocity. The command now uses Bukkit/Paper entity effect APIs without loading PacketEvents classes.
- **Changes** - Removed direct PacketEvents usage from `TotemAnimationCommand`.

### 🟠 fix: Recipe Result Validation
- **Summary** - Prevent invalid recipe outputs from being passed into Bukkit recipe constructors.
- **Description** - Recipe configurations with invalid or missing outputs could result in null `ItemStack` results being silently passed into Bukkit recipe constructors, causing errors that were difficult to trace back to the misconfigured recipe.
- **Changes** - Added shared recipe result validation so invalid recipe outputs are reported before null `ItemStack` results are passed into Bukkit recipe constructors.

### 🟠 feat: Loaded Item State Updates
- **Summary** - Update loaded item state beyond player inventories so existing in-world and loaded container items are refreshed safely.
- **Description** - Oraxen now applies the existing item update routine to loaded entities, entity inventories and equipment, player ender chests, and loaded tile-entity inventories. Startup, entity load, chunk load, player join, and item reload paths now cover loaded server state without editing offline playerdata from disk.

- **Changes** - Added loaded entity and tile-entity updater gates, scheduled startup and chunk/entity scans through existing Folia-safe scheduler helpers, reused `ItemUpdater.updateItem` for all touched inventories, and updated reload handling to refresh ender chests plus loaded entity contents.

### 🟠 chore: TriumphGUI Inventory Migration
- **Summary** - Replace Oraxen’s remaining InventoryFramework GUI usage with TriumphGUI to improve Oraxen jar size.
- **Description** - The recipe showcase inventory now uses TriumphGUI while preserving its existing 6-row layout, recipe/result slots, close button, previous/next navigation, click cancellation, drag cancellation, and outside-click handling. InventoryFramework is no longer declared in the shaded dependency bundle or documentation. **This improved the Oraxen `.jar` size from 5.1MB to 3.2MB.**

- **Changes** - Migrated `RecipesView`, updated recipe showcase opening through `InvManager`/`RecipesCommand`, removed the InventoryFramework version catalog entry from `libraries-shade`, and cleaned obsolete InventoryFramework documentation references.

### 🟠 chore: CodeFactor NMS Handler Exclusions
- **Summary** - Exclude version-specific NMS handler adapters from CodeFactor analysis.
- **Description** - CodeFactor’s remaining C-grade files were version-bound `NMSHandler` adapters whose duplication comes from mirroring Minecraft internals across supported server versions. The furniture file already shows no active CodeFactor issues on its detail page, so behavior-sensitive gameplay code was left unchanged.

- **Changes** - Added `.codefactor.yml` with an exclude pattern for versioned `NMSHandler.java` files so the intentional adapter duplication no longer drives repository grades.

### 🟠 chore: Stop Gestures Folder Generation
- **Summary** - Stop creating the unused `plugins/Oraxen/gestures` folder during Oraxen startup.
- **Description** - Oraxen previously initialized a `gestures` data folder and attempted to extract bundled gesture YAML configs when default config generation was enabled. No gesture resources are bundled in the current resource tree, so this startup path only produced an empty, unused folder.

- **Changes** - Removed the `gesturesFolder` field and the `validatesConfig()` block that created and populated `plugins/Oraxen/gestures`.

### 🟠 feat: Lobfile Pack Hosting
- **Summary** - Add Lobfile as a built-in pack host with support for single and multi-version resource pack uploads.
- **Description** - Oraxen can now use `Pack.upload.type: lobfile` with `Pack.upload.lobfile.api-key` and `Pack.upload.lobfile.pack-name`. Single-pack uploads use the configured pack name, while multi-version uploads name each Lobfile upload with `<lobfile.pack-name>_<version>` so multiple generated packs can coexist.

- **Changes** - Added a Lobfile hosting provider for `https://lobfile.com/api/v3/upload`, wired it into the hosting provider factory and default settings, passed version context through multi-version uploads, refreshed multi-version configuration guidance, and added tests for Lobfile upload filename generation.

### 🟠 feat: Custom Paintings
- **Summary** - Added custom painting support through generated datapack entries and `Components.painting_variant` item configuration.
- **Description** - Oraxen now loads `paintings.yml`, generates an `oraxen_paintings` datapack with painting variant JSON and optional random-placement tags, and applies painting variants to painting items across supported 1.21 server versions. The shared datapack writer now creates its target folder before writing `pack.mcmeta`.

- **Changes** - Added `CustomPainting` and `PaintingDatapack`, registered `paintings.yml` in config validation, wired painting datapack generation into resource-pack generation, added datapack format resolution for server data, added `painting_variant` item parsing/application with Paper API support on newer versions and NMS fallback for 1.21-1.21.4, and included a commented example painting item config.

### 🟠 ultra-feature: Unified Custom Block Mechanic Refactor
- **Summary** - Replaced Oraxen’s fragmented custom block mechanics with a single unified `block` mechanic and a shared breaking/drop architecture.
- **Description** - The old deprecated `block` mechanic was removed, and the former `noteblock`, `stringblock`, `chorusblock`, and `shaped_block` mechanic paths were folded behind one root `Mechanics.block` entry. Block type is now selected with `type`, covering `FULL`, `STRING`, `CHORUS`, `STAIR`, `SLAB`, `DOOR`, `TRAPDOOR`, `GRATE`, and `BULB`.

- **Changes** - Added unified block factory dispatch, unified custom block lookup/removal APIs, shaped block persistence under the shared block key, shared light/storage cleanup, shared break sound handling, shaped block mining support, unified resource-pack modifier registration, updated metrics/schema/default resources, and removed old deprecated block listener/mechanic code. Breaking now uses ordered `breaking` rules with per-rule hardness and drops, supports tool IDs and item tags, and correctly compensates for backing-block vanilla mining speed on fallback rules.
This fixes around 20 bugs that the previous mechanics had.

```yaml
Mechanics:
    block:
      type: FULL # FULL, STRING, CHORUS, STAIR, SLAB, DOOR, TRAPDOOR, GRATE, BULB
      block-sounds:
        break-sound: block.wood.break
        place-sound: block.wood.place
      custom-variation: 5
      appearance: # Also supports textures
        model: crystalmush_log
      breaking: # New drop & breaking config section
      - when: # When the tool is...
        - minecraft:iron_axe
        - minecraft:golden_axe
        - minecraft:diamond_axe
        - minecraft:netherite_axe # Also supports tags via "#<tag>"
        hardness: 2 # Hardness when these tools are used
        drops: # Drops when this tool is used
        - item: crystalmush_log
          probability: 1.0
          silk-touch: false # Only drop when this tool has silk touch
          fortune: 1 # Percentage (1 = 100%) to increase the loot amount per fortune level
      - else: # Anything that is not covered by another `when` section
        hardness: 4
        drops:
        - item: crystalmush_log
          probability: 1.0
```

### 🟠 fix: Totem Animation Command API Loading
- **Summary** - Prevents `/oraxen totem-animation` command registration from crashing servers that do not provide Paper's newer data component API.
- **Description** - `TotemAnimationCommand` previously imported `io.papermc.paper.datacomponent.*` classes directly, which caused `NoClassDefFoundError` during plugin enable on Paper 1.20.4 and likely other unsupported runtimes before the command was ever executed. The command now gates death-protection component handling behind Paper + 1.21.2+ checks and resolves the Paper API reflectively only when supported.

- **Changes** - Removed direct Paper data-component imports from `TotemAnimationCommand`, added reflective `setData`/`hasData` calls for `DEATH_PROTECTION`, and kept unsupported servers on the existing totem/material fallback path.

### 🟠 fix: Gate Generated Shaders For Legacy Packs
- **Summary** - Excludes Oraxen-generated core text shaders from final resource-pack outputs targeting Minecraft versions below 1.21.4.
- **Description** - The shader generation pipeline still generates text-effect and animated-glyph shaders where needed, but final pack selection now filters generated shader outputs for legacy targets. The filtering is path and SHA-256 based so externally merged shader content from uploads/assets is not removed unless it exactly matches a generated shader from the current run.

- **Changes** - Tracks generated core shader hashes in `TextShaderGenerator`; passes generated shader metadata into `MultiVersionPackGenerator`; filters matching generated shaders from multi-version pack outputs below 1.21.4; applies the same guard to single-pack output for legacy servers.


pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockBreaking.java
```diff
@@ -0,0 +1,243 @@
...
+    private final List<Rule> rules;
+
+    public BlockBreaking(ConfigurationSection section, String sourceID) {
+        this.rules = parseRules(section.getList("breaking"), sourceID);
+    }
+
+    public boolean hasHardness(ItemStack tool) {
+        Rule rule = ruleFor(tool);
+        return rule != null && rule.hardness() >= 0.0D;
+    }
```
miziusBot, line 32: <a href="#">P2</a> When a block has a configured `breaking` rule with `hardness: 0`, `hasHardness` returns `true` (because `0.0 >= 0.0`). In the shaped-block HardnessModifier, `getPeriod` computes `Math.round(0 * 0.4 / speed) == 0` and then the `period == 0 && mechanic.hasHardness(tool)` guard correctly returns `1`. However, for vanilla block mechanic usages that go through `attributeSpeedMultiplier`, `configuredToolSpeed` is `tierToolSpeed(tool, blockType)` and `nativeToolSpeed` is `tierToolSpeed` too for matching tools, giving a multiplier of 1.0 — the player mines at vanilla speed regardless of the intended `hardness: 0` config. Was `hardness: 0` intended to mean instant-break, or "use default vanilla speed"?

```suggestion
    public boolean hasHardness(ItemStack tool) {
        Rule rule = ruleFor(tool);
        return rule != null && rule.hardness() > 0.0D;
    }
```

core/src/main/java/io/th0rgal/oraxen/pack/generation/MultiVersionPackGenerator.java
```diff
@@ -246,6 +264,27 @@ private boolean isPackMcmeta(String path) {
...
+            return false;
+        }
+
+        if (minecraftVersion.isAtLeast(GENERATED_CORE_SHADER_MIN_VERSION)) {
+            return false;
+        }
+
+        String generatedHash = generatedCoreShaderHashes.get(path);
+        return generatedHash != null && generatedHash.equals(sha256(content));
+    }
```
miziusBot, line 282: <a href="#">P2</a> **Overlay shader paths incorrectly excluded from legacy packs**

`shouldExcludeGeneratedCoreShader` is applied to all `mFile.path` values, including overlay-directory paths such as `overlay_v1_21_4/assets/minecraft/shaders/core/rendertype_text.vsh`. For a pack targeting a legacy Minecraft version the condition also filters those overlay shader files, because their hash-map key matches and `minecraftVersion.isAtLeast(1.21.4)` is false. A 1.20.2+ client that consumes overlays from the legacy-target ZIP (e.g. via ViaVersion) would silently lose those shaders. The exclusion filter should only apply to base-path shader files (no overlay-directory prefix).

pxlarified: ✅ Fixed in 7e0df178abd6abcdaadd020ba5a2c2232bc69805.

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockBreaking.java
pxlarified: ✅ Fixed in 7e0df178abd6abcdaadd020ba5a2c2232bc69805.

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
```diff
@@ -144,6 +213,105 @@ public void onUseConvertedTo(PlayerItemConsumeEvent event) {
...
+
+        SchedulerUtil.runTask(() -> {
+            for (World world : Bukkit.getWorlds()) {
+                for (Entity entity : world.getEntities()) {
+                    if (!shouldUpdateEntityContents(entity)) continue;
+                    SchedulerUtil.runForEntity(entity, () -> updateEntityInventories(entity), () -> {});
+                }
+            }
+        });
+    }
```
miziusBot, line 233: <a href="#">P2</a> **Folia-unsafe world access in global task context**

`world.getEntities()` is called inside `SchedulerUtil.runTask()`. On Folia, the global task scheduler does not own region data, so iterating `world.getEntities()` from it is undefined behaviour — entities belong to region threads and must be accessed from the owning region. The same concern applies to `world.getLoadedChunks()` in `updateLoadedTileEntityContents()`. Both methods are now public and called from `ReloadCommand` on a command-handler thread, compounding the issue. If `SchedulerUtil.runTask()` routes to Folia's global scheduler, the entity/chunk scan should be moved to individual region schedulers.

core/src/main/java/io/th0rgal/oraxen/pack/upload/hosts/Lobfile.java
```diff
@@ -0,0 +1,195 @@
...
+    private static final String DEFAULT_PACK_NAME = "Oraxen";
+    private static final int CONNECT_TIMEOUT_MS = 5_000;
+    private static final int CONNECTION_REQUEST_TIMEOUT_MS = 5_000;
+    private static final int SOCKET_TIMEOUT_MS = 30_000;
+
+    private final String apiKey;
+    private final String packName;
+    private String packUrl;
+    private String sha1;
+    private UUID packUUID;
```
miziusBot, line 42: <a href="#">P2</a> **Instance-level upload state shared across multi-version calls**

`packUrl`, `sha1`, and `packUUID` are mutable instance fields overwritten by each `uploadPackWithName` call. In the multi-version flow, `MultiVersionUploadManager` reads `getPackURL()` / `getSHA1()` / `getPackUUID()` after each upload using the same `Lobfile` instance. This is safe only if reads happen synchronously after each individual upload before the next begins. If callers are ever refactored to read these after all uploads finish, they will silently receive only the last version's data.

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockBreaking.java
```diff
@@ -0,0 +1,243 @@
...
+
+    private double parseDouble(Object value, double fallback, String sourceID) {
+        if (value == null) return fallback;
+        try {
+            return Double.parseDouble(value.toString());
+        } catch (NumberFormatException exception) {
+            throw new IllegalArgumentException("Invalid hardness value '" + value + "' in block mechanic " + sourceID, exception);
+        }
+    }
```
miziusBot, line 170: <a href="#">P2</a> **Invalid hardness throws unchecked exception during mechanic parsing**

`parseDouble` throws `IllegalArgumentException` on a bad `hardness` value, propagating out of `BlockBreaking`'s constructor. If `BlockMechanicFactory.parse()` or its delegate factories do not catch this, a single misconfigured block item can abort mechanic loading entirely without a user-visible warning. Consider logging the bad value and returning a default hardness instead, consistent with how other config parse errors are handled.

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
pxlarified: Fixed in 1ac24ec6 by avoiding reload-time full-world entity/chunk scans on Folia; entity and tile-entity contents continue to be updated from their load events on the owning scheduler.

core/src/main/java/io/th0rgal/oraxen/pack/upload/hosts/Lobfile.java
pxlarified: Fixed in 1ac24ec6 by creating a fresh hosting provider per pack-version upload, so mutable provider upload state is no longer shared across multi-version uploads.

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockBreaking.java
pxlarified: Fixed in 1ac24ec6 by logging invalid hardness values and falling back to the default hardness instead of throwing during mechanic parsing.

pxlarified: ✅ Fixed in 1ac24ec63765a5ee2151eae2c03c14470393992a.

core/src/main/java/io/th0rgal/oraxen/pack/upload/hosts/Lobfile.java
pxlarified: ✅ Fixed in 1ac24ec63765a5ee2151eae2c03c14470393992a.

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
pxlarified: ✅ Fixed in 1ac24ec63765a5ee2151eae2c03c14470393992a.

PR #1834, opened 2026-05-25
Oraxen v1.215.0

pxlarified:
### 🟠 fix: Folia Shaped Block Chunk Conversion
- **Summary** - Schedule shaped-block waxed copper chunk conversion on the chunk's region thread instead of the Folia global scheduler.
- **Description** - Newly generated chunks can trigger waxed copper conversion after load. On Folia 1.21.11, running that delayed conversion globally can read block state from the wrong context and hit a null `Level.getCurrentWorldData()` path. The conversion now uses `SchedulerUtil.runAtLocationLater` with the chunk location and skips work if the chunk unloads before the delayed task runs.

- **Changes** - Updated `ShapedBlockMechanicListener#onChunkLoad` to dispatch conversion through the region scheduler, and added an `isLoaded` guard before scanning chunk blocks.

### 🟠 fix: Optional PacketEvents Startup
- **Summary** - Keep Oraxen command registration from hard-requiring PacketEvents when the optional plugin is not installed.
- **Description** - Oraxen 1.214.0 added the totem animation command with direct PacketEvents class references. Servers without PacketEvents could throw `NoClassDefFoundError: com/github/retrooper/packetevents/wrapper/PacketWrapper` while enabling Oraxen, disabling the plugin and preventing players from joining through setups such as Velocity. The command now uses Bukkit/Paper entity effect APIs without loading PacketEvents classes.
- **Changes** - Removed direct PacketEvents usage from `TotemAnimationCommand`.

### 🟠 fix: Recipe Result Validation
- **Summary** - Prevent invalid recipe outputs from being passed into Bukkit recipe constructors.
- **Description** - Recipe configurations with invalid or missing outputs could result in null `ItemStack` results being silently passed into Bukkit recipe constructors, causing errors that were difficult to trace back to the misconfigured recipe.
- **Changes** - Added shared recipe result validation so invalid recipe outputs are reported before null `ItemStack` results are passed into Bukkit recipe constructors.

### 🟠 feat: Loaded Item State Updates
- **Summary** - Update loaded item state beyond player inventories so existing in-world and loaded container items are refreshed safely.
- **Description** - Oraxen now applies the existing item update routine to loaded entities, entity inventories and equipment, player ender chests, and loaded tile-entity inventories. Startup, entity load, chunk load, player join, and item reload paths now cover loaded server state without editing offline playerdata from disk.

- **Changes** - Added loaded entity and tile-entity updater gates, scheduled startup and chunk/entity scans through existing Folia-safe scheduler helpers, reused `ItemUpdater.updateItem` for all touched inventories, and updated reload handling to refresh ender chests plus loaded entity contents.

### 🟠 chore: TriumphGUI Inventory Migration
- **Summary** - Replace Oraxen’s remaining InventoryFramework GUI usage with TriumphGUI to improve Oraxen jar size.
- **Description** - The recipe showcase inventory now uses TriumphGUI while preserving its existing 6-row layout, recipe/result slots, close button, previous/next navigation, click cancellation, drag cancellation, and outside-click handling. InventoryFramework is no longer declared in the shaded dependency bundle or documentation. **This improved the Oraxen `.jar` size from 5.1MB to 3.2MB.**

- **Changes** - Migrated `RecipesView`, updated recipe showcase opening through `InvManager`/`RecipesCommand`, removed the InventoryFramework version catalog entry from `libraries-shade`, and cleaned obsolete InventoryFramework documentation references.

### 🟠 chore: CodeFactor NMS Handler Exclusions
- **Summary** - Exclude version-specific NMS handler adapters from CodeFactor analysis.
- **Description** - CodeFactor’s remaining C-grade files were version-bound `NMSHandler` adapters whose duplication comes from mirroring Minecraft internals across supported server versions. The furniture file already shows no active CodeFactor issues on its detail page, so behavior-sensitive gameplay code was left unchanged.

- **Changes** - Added `.codefactor.yml` with an exclude pattern for versioned `NMSHandler.java` files so the intentional adapter duplication no longer drives repository grades.

### 🟠 chore: Stop Gestures Folder Generation
- **Summary** - Stop creating the unused `plugins/Oraxen/gestures` folder during Oraxen startup.
- **Description** - Oraxen previously initialized a `gestures` data folder and attempted to extract bundled gesture YAML configs when default config generation was enabled. No gesture resources are bundled in the current resource tree, so this startup path only produced an empty, unused folder.

- **Changes** - Removed the `gesturesFolder` field and the `validatesConfig()` block that created and populated `plugins/Oraxen/gestures`.

### 🟠 feat: Lobfile Pack Hosting
- **Summary** - Add Lobfile as a built-in pack host with support for single and multi-version resource pack uploads.
- **Description** - Oraxen can now use `Pack.upload.type: lobfile` with `Pack.upload.lobfile.api-key` and `Pack.upload.lobfile.pack-name`. Single-pack uploads use the configured pack name, while multi-version uploads name each Lobfile upload with `<lobfile.pack-name>_<version>` so multiple generated packs can coexist.

- **Changes** - Added a Lobfile hosting provider for `https://lobfile.com/api/v3/upload`, wired it into the hosting provider factory and default settings, passed version context through multi-version uploads, refreshed multi-version configuration guidance, and added tests for Lobfile upload filename generation.

### 🟠 feat: Custom Paintings
- **Summary** - Added custom painting support through generated datapack entries and `Components.painting_variant` item configuration.
- **Description** - Oraxen now loads `paintings.yml`, generates an `oraxen_paintings` datapack with painting variant JSON and optional random-placement tags, and applies painting variants to painting items across supported 1.21 server versions. The shared datapack writer now creates its target folder before writing `pack.mcmeta`.

- **Changes** - Added `CustomPainting` and `PaintingDatapack`, registered `paintings.yml` in config validation, wired painting datapack generation into resource-pack generation, added datapack format resolution for server data, added `painting_variant` item parsing/application with Paper API support on newer versions and NMS fallback for 1.21-1.21.4, and included a commented example painting item config.

### 🟠 ultra-feature: Unified Custom Block Mechanic Refactor
- **Summary** - Replaced Oraxen’s fragmented custom block mechanics with a single unified `block` mechanic and a shared breaking/drop architecture.
- **Description** - The old deprecated `block` mechanic was removed, and the former `noteblock`, `stringblock`, `chorusblock`, and `shaped_block` mechanic paths were folded behind one root `Mechanics.block` entry. Block type is now selected with `type`, covering `FULL`, `STRING`, `CHORUS`, `STAIR`, `SLAB`, `DOOR`, `TRAPDOOR`, `GRATE`, and `BULB`.

- **Changes** - Added unified block factory dispatch, unified custom block lookup/removal APIs, shaped block persistence under the shared block key, shared light/storage cleanup, shared break sound handling, shaped block mining support, unified resource-pack modifier registration, updated metrics/schema/default resources, and removed old deprecated block listener/mechanic code. Breaking now uses ordered `breaking` rules with per-rule hardness and drops, supports tool IDs and item tags, and correctly compensates for backing-block vanilla mining speed on fallback rules.
This fixes around 20 bugs that the previous mechanics had.

```yaml
Mechanics:
    block:
      type: FULL # FULL, STRING, CHORUS, STAIR, SLAB, DOOR, TRAPDOOR, GRATE, BULB
      block-sounds:
        break-sound: block.wood.break
        place-sound: block.wood.place
      custom-variation: 5
      appearance: # Also supports textures
        model: crystalmush_log
      breaking: # New drop & breaking config section
      - when: # When the tool is...
        - minecraft:iron_axe
        - minecraft:golden_axe
        - minecraft:diamond_axe
        - minecraft:netherite_axe # Also supports tags via "#<tag>"
        hardness: 2 # Hardness when these tools are used
        drops: # Drops when this tool is used
        - item: crystalmush_log
          probability: 1.0
          silk-touch: false # Only drop when this tool has silk touch
          fortune: 1 # Percentage (1 = 100%) to increase the loot amount per fortune level
      - else: # Anything that is not covered by another `when` section
        hardness: 4
        drops:
        - item: crystalmush_log
          probability: 1.0
```

### 🟠 fix: Totem Animation Command API Loading
- **Summary** - Prevents `/oraxen totem-animation` command registration from crashing servers that do not provide Paper's newer data component API.
- **Description** - `TotemAnimationCommand` previously imported `io.papermc.paper.datacomponent.*` classes directly, which caused `NoClassDefFoundError` during plugin enable on Paper 1.20.4 and likely other unsupported runtimes before the command was ever executed. The command now gates death-protection component handling behind Paper + 1.21.2+ checks and resolves the Paper API reflectively only when supported.

- **Changes** - Removed direct Paper data-component imports from `TotemAnimationCommand`, added reflective `setData`/`hasData` calls for `DEATH_PROTECTION`, and kept unsupported servers on the existing totem/material fallback path.

### 🟠 fix: Gate Generated Shaders For Legacy Packs
- **Summary** - Excludes Oraxen-generated core text shaders from final resource-pack outputs targeting Minecraft versions below 1.21.4.
- **Description** - The shader generation pipeline still generates text-effect and animated-glyph shaders where needed, but final pack selection now filters generated shader outputs for legacy targets. The filtering is path and SHA-256 based so externally merged shader content from uploads/assets is not removed unless it exactly matches a generated shader from the current run.

- **Changes** - Tracks generated core shader hashes in `TextShaderGenerator`; passes generated shader metadata into `MultiVersionPackGenerator`; filters matching generated shaders from multi-version pack outputs below 1.21.4; applies the same guard to single-pack output for legacy servers.


pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/utils/drops/Loot.java
```diff
@@ -112,25 +150,56 @@ public int getMaxAmount() {
...
+        if (Math.random() < exactAmount - roundedAmount) roundedAmount++;
+        return Math.max(1, roundedAmount);
+    }
+
     private void dropItems(Location location, int amountMultiplier) {
         ItemStack item = getItem(amountMultiplier);
         if (location.getWorld() != null && item != null) location.getWorld().dropItemNaturally(location, item);
     }
 
+    private void dropItems(Location location, int amountMultiplier, ItemStack tool) {
```
miziusBot: <a href="#">P2</a> The no-arg-tool overload of `dropNaturally` now delegates to the three-argument form, making `dropItems(Location, int)` unreachable dead code. It can be removed to avoid confusion about which path actually executes.

```suggestion
    private void dropItems(Location location, int amountMultiplier, ItemStack tool) {
```

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
```diff
@@ -144,6 +214,113 @@ public void onUseConvertedTo(PlayerItemConsumeEvent event) {
...
+    private static void updateTileEntityInventories(Chunk chunk) {
+        for (BlockState tileEntity : chunk.getTileEntities()) {
+            if (!TILE_ENTITY_TYPES.contains(tileEntity.getType()) || !(tileEntity instanceof InventoryHolder holder)) continue;
+            updateInventory(holder.getInventory());
+        }
+    }
+
+    public static void updateEntityInventories(Entity entity) {
+        if (entity instanceof ItemFrame itemFrame) {
+            ItemStack oldItem = itemFrame.getItem();
```
miziusBot, line 311: <a href="#">P2</a> **Double-update for entities that implement both interfaces**

`updateEntityInventories` calls `updateInventory(holder.getInventory())` and then `updateEquipment(livingEntity)` as two separate `if` blocks (not `else if`). For entities that implement both `InventoryHolder` and `LivingEntity` — such as `AbstractHorse` subclasses, where `getInventory()` exposes the saddle/armor slots — the equipment is processed twice. The second pass will find the already-updated item unchanged and skip the write, so there is no data corruption, but the redundant slot scan happens on every entity load and startup sweep. Consider guarding the `updateEquipment` branch with `else if (entity instanceof LivingEntity)` so it only runs when the entity wasn't already handled as an `InventoryHolder`.

pxlarified: ✅ Fixed in 05d1bfce5bd7690cca1126eace3d0ddcf12e0be5.

core/src/main/java/io/th0rgal/oraxen/utils/drops/Loot.java
pxlarified: ✅ Fixed in 05d1bfce5bd7690cca1126eace3d0ddcf12e0be5.

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
```diff
@@ -144,6 +214,113 @@ public void onUseConvertedTo(PlayerItemConsumeEvent event) {
...
+            if (!Objects.equals(oldItem, newItem)) itemDisplay.setItemStack(newItem);
+        }
+        if (entity instanceof Item item) {
+            ItemStack oldItem = item.getItemStack();
+            ItemStack newItem = updateItem(oldItem);
+            if (!Objects.equals(oldItem, newItem)) item.setItemStack(newItem);
+        }
+        if (entity instanceof InventoryHolder holder) updateInventory(holder.getInventory());
+        else if (entity instanceof LivingEntity livingEntity) updateEquipment(livingEntity);
+    }
```
miziusBot, line 327: <a href="#">P2</a> **Equipment not updated for `InventoryHolder` entities that are also `LivingEntity`**

The `if / else if` means that for any entity that is both `InventoryHolder` and `LivingEntity` — Horses, Donkeys, Mules, and Llamas — only `updateInventory(holder.getInventory())` runs; `updateEquipment` is never called. Saddle and armor slots on those entity types will therefore not be refreshed during startup or entity-load scans. Changing the `else if` to a plain `if` (with the same Player exclusion already in `shouldUpdateEntityContents`) would cover both paths.

pxlarified: ✅ Fixed in 16ce7b9aaf95fb4e8bbc54f9b7e89abafba66d8b.

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
```diff
@@ -26,30 +38,88 @@
...
+import org.bukkit.inventory.meta.Damageable;
+import org.bukkit.inventory.meta.ItemMeta;
+import org.bukkit.inventory.meta.LeatherArmorMeta;
+import org.bukkit.inventory.meta.MapMeta;
+import org.bukkit.inventory.meta.PotionMeta;
 import org.bukkit.persistence.PersistentDataContainer;
 
-import java.util.*;
+import java.util.EnumSet;
+import java.util.List;
```
miziusBot, line 58: <a href="#">P2</a> **Startup batch scan may schedule thousands of tasks at once**

`updateLoadedContents()` is called 2 ticks after `ItemUpdater` is constructed. On a server with many loaded chunks and entities, this schedules one deferred task per entity (via `runForEntity`) and one per chunk (via `runAtLocation`) — all submitted at the same time on plugin enable. On large or heavily-pre-loaded servers this burst of scheduling could cause a noticeable lag spike at startup. Consider batching the startup scan across several ticks, or spreading it with a fixed-rate task.

core/src/main/java/io/th0rgal/oraxen/utils/drops/Loot.java
```diff
@@ -23,13 +25,17 @@ public class Loot {
...
+    private final double fortuneBonus;
     private LinkedHashMap<String, Object> config;
 
     public Loot(LinkedHashMap<String, Object> config, String sourceID) {
-        this.probability = Double.parseDouble(config.getOrDefault("probability", 1).toString());
+        this.probability = parseDouble(config.get("probability"), 1.0D);
         if (config.getOrDefault("amount", "") instanceof String amount && amount.contains("..")) {
             this.amount = Utils.parseToRange(amount);
         } else this.amount = new IntegerRange(1,1);
+        this.requiresSilkTouch = Boolean.parseBoolean(config.getOrDefault("silk-touch", false).toString());
```
miziusBot, line 37: <a href="#">P2</a> **`fortune` key in loot config now silently changes existing drops**

The new `fortuneBonus` is parsed from `config.get("fortune")`. Before this PR, `fortune` at the loot-entry level was read from config but never applied. Any existing noteblock/stringblock/chorusblock YAML loot sections that already contained `fortune: <value>` (e.g., copied from examples) will now have that bonus activated without any migration notice. If the parent `Drop` also has `fortune: true`, fortune would effectively be applied twice — once by `Drop.getFortuneMultiplier` passed as `amountMultiplier`, and once by `Loot.applyFortune`. This is unlikely in practice but worth documenting in a migration note.

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
```diff
@@ -258,7 +435,7 @@ public static ItemStack updateItem(ItemStack oldItem) {
 
             if (VersionUtil.atOrAbove("1.21.2")) {
                 if (newMeta.hasEquippable()) itemMeta.setEquippable(newMeta.getEquippable());
-                else if (oldMeta.hasEquippable()) itemMeta.setEquippable(newMeta.getEquippable());
+                else if (oldMeta.hasEquippable()) itemMeta.setEquippable(oldMeta.getEquippable());
```
miziusBot, line 478: <a href="#">P2</a> **Equippable fallback should preserve `oldMeta`'s equippable data** — this looks like a pre-existing bug that this PR correctly fixes: the `else if` branch previously called `newMeta.getEquippable()` instead of `oldMeta.getEquippable()`, which would lose the original equippable data when the new item template has no equippable component. The fix is already in the PR; leaving this as a positive note to make the intent explicit in a comment.

```suggestion
                // Preserve old item's equippable data when the new template has none.
                else if (oldMeta.hasEquippable()) itemMeta.setEquippable(oldMeta.getEquippable());
```

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockBreaking.java
```diff
@@ -0,0 +1,245 @@
...
+        if (!(value instanceof List<?> dropConfigs)) return Drop.emptyDrop();
+
+        List<Loot> loots = new ArrayList<>();
+        for (Object entry : dropConfigs) {
+            if (!(entry instanceof Map<?, ?> map)) continue;
+            loots.add(new Loot(new LinkedHashMap<>((Map<String, Object>) map), sourceID));
+        }
+
+        return Drop.emptyDrop(loots);
+    }
```
miziusBot: <a href="#">P2</a> **Unchecked cast from `Map<?, ?>` to `Map<String, Object>` could throw `ClassCastException`**

SnakeYAML parses integer-keyed YAML entries (e.g. `1:`) as `Integer`, not `String`. If a `drops` block entry uses integer keys the cast silently compiles but fails at runtime inside `LinkedHashMap`'s copy constructor. Explicitly converting keys to `String` is safer and removes the need for `@SuppressWarnings`.

```suggestion
    private Drop parseDrop(Object value, String sourceID) {
        if (!(value instanceof List<?> dropConfigs)) return Drop.emptyDrop();

        List<Loot> loots = new ArrayList<>();
        for (Object entry : dropConfigs) {
            if (!(entry instanceof Map<?, ?> map)) continue;
            LinkedHashMap<String, Object> safeMap = new LinkedHashMap<>();
            for (Map.Entry<?, ?> e : map.entrySet()) safeMap.put(String.valueOf(e.getKey()), e.getValue());
            loots.add(new Loot(safeMap, sourceID));
        }

        return Drop.emptyDrop(loots);
    }
```

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
pxlarified: ✅ Fixed in 29171ba5.

core/src/main/java/io/th0rgal/oraxen/utils/drops/Loot.java
pxlarified: ✅ Fixed in 29171ba5.

core/src/main/java/io/th0rgal/oraxen/items/ItemUpdater.java
pxlarified: ✅ Fixed in 29171ba5.

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockBreaking.java
pxlarified: ✅ Fixed in 29171ba5.

PR #1838, opened 2026-06-04
v1.217.0

pxlarified:
## Oraxen v1.217.0

This pull request is not ready to be merged.

This update focuses on improvements, chores and cleanups instead of new features.

## Fixes

- **Fixed** backpack recipe ingredient matching.
  - **Summary** - Fixed crafting recipes that use Oraxen backpack items as ingredients after the backpack has stored item data.
  - **Description** - Backpack items now remain usable in configured Oraxen workbench recipes even after being opened or filled, while recipes still require the correct Oraxen item ID instead of accepting any item with the same base material.
  - **Changes** - Registered backpack Oraxen ingredients with material-based Bukkit choices for shaped and shapeless workbench recipes, tracked expected Oraxen ingredient IDs from recipe configs, validated crafting matrices during prepare craft events, and reset the tracked recipe data on recipe reload.

- **Persisted** backpack contents in crafting.
  - **Summary** - Persisted stored backpack contents when crafting backpack mechanic items from existing backpacks.
  - **Description** - Backpack-to-backpack crafting now carries the source backpack inventory data into the crafted result, preventing stored items from being lost when recipes upgrade or transform backpacks.
  - **Changes** - Added prepare-craft result handling that detects backpack mechanic results, copies the backpack persistent data container contents from a backpack ingredient, and reapplies the persisted data to both whitelisted and rewritten Oraxen crafting results.

- **Fixed** BlockLocker furniture protection.

  - **Summary** - Fixed BlockLocker integration so only furniture is registered as protectable and Oraxen furniture interactions respect BlockLocker access checks.

  - **Description** - Furniture configured with the `blocklocker` mechanic now remains the only Oraxen mechanic integrated with BlockLocker, and players denied by BlockLocker can no longer interact with protected furniture through Oraxen's furniture interaction handlers.

  - **Changes** - Removed BlockLocker parsing from note, string, and chorus block mechanics, limited BlockLocker protectable lookup to furniture, de-duplicated BlockLocker protectable registration, added a BlockLocker permission helper, and guarded furniture entity and barrier interaction paths with `BlockLockerAPIv2.isAllowed`.

- **Fixed** player head item definition fallback.

  - **Summary** - Fixed 1.21.4+ vanilla item definition fallbacks for player heads.

  - **Description** - Normal player heads now keep Minecraft's specialized skull renderer when Oraxen adds custom player-head item dispatches, preventing the vanilla fallback from rendering through the generic model path while preserving custom Oraxen item cases.

  - **Changes** - Added special skull/head fallback generation for `PLAYER_HEAD`, included obvious vanilla skull `kind` mappings, preserved select, range_dispatch, and combined dispatch behavior, and added focused `VanillaItemDefinitionGeneratorTest` coverage.

- **Fixed** glyph shadow color handling.

  - **Summary** - Fixed glyph `appearance.shadow_color` application for supported Adventure runtimes.

  - **Description** - Oraxen now applies glyph shadow colors based on runtime API support instead of a server-version gate, logs a warning when shadow colors are configured but unavailable, and avoids wrapping glyph components in an extra parent component that could interfere with shadow color rendering.

  - **Changes** - Removed the hard 1.21.4 version check from `GlyphAppearance`, added one-time unavailable-shadow warnings, returned glyph and reference glyph components directly after applying shadow color, and updated modern chat raw-glyph replacement to preserve configured glyph appearance.

- **Fixed** item tool component version handling.

  - **Summary** - Guarded item tool component reads and writes behind Minecraft 1.20.6+ checks.

  - **Description** - Oraxen now avoids accessing item tool component APIs on unsupported versions, preventing compatibility issues when loading items on older Paper versions.

  - **Changes** - Added version checks around `ItemMeta` tool component handling in `ItemBuilder`.

- **Validated** `/oraxen give` amounts.

  - **Summary** - Added validation for `/oraxen give` amount input.

  - **Description** - The give command now rejects zero and negative amounts before processing, preventing invalid input from crashing command execution while preserving normal behavior for valid positive amounts.

  - **Changes** - Added positive amount validation, introduced an invalid amount language message, fixed large amount stack cap calculation, and added `CommandsManagerTest` coverage for invalid amounts and stack splitting.

- **Guarded** manager event cleanup.

  - **Summary** - Made font and HUD manager replacement safe when no previous manager exists.

  - **Description** - `setFontManager` and `setHudManager` now tolerate partial startup, reload, and test paths where the previous manager may be uninitialized, avoiding null dereferences while still unregistering existing managers and registering replacements.

  - **Changes** - Added null guards around previous manager event unregistration, guarded replacement registration, and added `OraxenPluginTest` coverage for font and HUD manager replacement paths.

- **Fixed** MiniMessage parsing for custom paintings.

  - **Summary** - Fixed custom painting title and author formatting from `paintings.yml`.

  - **Description** - Custom painting tooltip text now parses MiniMessage formatting correctly instead of showing raw tags like `<red>`.

  - **Changes** - Updated runtime painting component conversion to deserialize MiniMessage before creating vanilla components, added a legacy-formatted fallback, and documented MiniMessage support in `paintings.yml`.

- **Fixed** unused block subtype activation.

  - **Summary** - Disabled unused unified block sub-mechanics when no blocks of that backing type are configured.

  - **Description** - The unified `block` mechanic no longer activates Noteblock, Stringblock, Chorusblock, or shaped-block handling unless at least one corresponding `type` is parsed, allowing servers to use one block subtype without changing vanilla behavior for the others.

  - **Changes** - Deferred delegate block factory listener and resource-pack modifier registration until first use, made subtype `isEnabled()` checks depend on parsed mechanics, and guarded shaped-block API lookups behind the shaped factory enabled state.

- **Fixed** Folia custom sound world mismatch.

  - **Summary** - Fixed Folia world mismatch errors in custom step and fall sound handling.

  - **Description** - Generic game events can reference a different world than the entity's current location after cross-world state changes such as death or respawn. Oraxen now skips custom block sound lookup when the event location and entity location no longer match, preventing Folia from throwing `IllegalStateException: World mismatch` while querying the standing block.

  - **Changes** - Validated `GenericGameEvent` and entity locations against the same loaded world in furniture and note-block sound listeners, and reused the captured entity location when playing player step/fall sounds.

- **Guarded** shaped block mechanic version support.

  - **Summary** - Guarded shaped block mechanic registration behind Minecraft 1.20.4+.

  - **Description** - Oraxen now skips shaped block initialization and parsing on older servers before touching copper door, trapdoor, grate, and bulb `Material` constants that do not exist on 1.20.1, preventing startup `NoSuchFieldError`s.

  - **Changes** - Added a 1.20.4+ support check to `ShapedBlockMechanicFactory`, returned early before `ShapedBlockType` initialization on unsupported versions, logged ignored shaped block configs, and added Paper 1.20.1 to version loading coverage.

## Features / Improvements

- **Added** NMS-only custom painting hot reload.

  - **Summary** - Added Paper 1.21.5+ NMS registry injection for custom paintings and removed painting datapack generation.

  - **Description** - Custom paintings now hot reload through the live painting variant registry, allowing changes in `paintings.yml` to apply without requiring a restart. Painting support is restricted to Paper 1.21.5+ and unavailable on Spigot.

  - **Changes** - Added `CustomPaintingRegistry`, wired startup/config/all/paintings reload flows, added `/oraxen reload paintings`, removed painting datapack generation from resource-pack generation, restricted painting item components to Paper 1.21.5+, added shared painting config parsing, updated `paintings.yml` docs, and disabled/removes the legacy paintings datapack during painting reload.

- **Added** version loading tests.

  - **Summary** - Added Paper Version loading tests that verify Oraxen loads fine on all supported minecraft versions.

  - **Description** - Downloads a server jar, cleans up the server, builds and deploys Oraxen to the server, sets the Java Version and verifies Oraxen loads fine on the server. Run via `./gradlew :core:test --tests io.th0rgal.oraxen.loading.VersionLoadingTest -PrunVersionLoadingTest=true` for all versions or `./gradlew :core:test --tests "io.th0rgal.oraxen.loading.VersionLoadingTest.Paper_26_1_2" -PrunVersionLoadingTest=true` for an individual version (e.g. `Paper_1_21_11`).

- **Added** configurable block break durability.

  - **Summary** - Added durability actions to block breaking rules.

  - **Description** - Block mechanics can now configure tool durability changes per matched breaking rule using `durability.add` and `durability.remove`, including fallback `else` rules.

  - **Changes** - Parsed durability actions in `BlockBreaking`, exposed them through note/string/chorus/shaped block mechanics, applied configured durability on successful block removal, prevented duplicate vanilla/custom damage, updated example block YAML configs, and added `BlockBreakingTest` coverage.

 

- **Migrated** to AntiGriefLib.

  - **Summary** - Migrated protection checks from ProtectionLib to AntiGriefLib.

  - **Description** - Oraxen protection checks for blocks, furniture, farming, combat, and misc mechanics now route through AntiGriefLib instead of ProtectionLib, adding support for 25+ plugins that ProtectionLib previously did not support.

  - **Changes** - Replaced the shaded ProtectionLib dependency with AntiGriefLib, added an Oraxen AntiGriefLib facade, migrated ProtectionLib imports/usages, updated README Maven exclusions, and added `AntiGriefLibTest` coverage.

- **Added** custom WorldGuard compatibility.

  - **Summary** - Added custom WorldGuard compatibility for Oraxen protection checks.

  - **Description** - Oraxen now registers its own WorldGuard compatibility through AntiGriefLib, allowing Oraxen actions to respect WorldGuard region flags and bypass permissions.

  - **Changes** - Added `OraxenWorldGuardCompatibility`, excluded AntiGriefLibÃƒÂ¢Ã¢â€šÂ¬Ã¢â€žÂ¢s default WorldGuard hook when WorldGuard is present, registered the custom compatibility, added WorldGuard dependency metadata, and included WorldGuard in the plugin dependency setup.

- **Added** support for copying bundled comments for new `settings.yml` options.

  - **Summary** - Added support for copying bundled settings.yml comments when inserting newly added settings into existing configs.

  - **Description** - Existing installs previously received only missing values, so newly added settings lacked the explanatory comments available in freshly generated configs. The updater now parses and preserves YAML comments, then copies related block and inline comments from the bundled default `settings.yml` for newly inserted options while leaving existing user values and comments intact. The Showcase can be found [here](<https://gist.github.com/pxlarified/f11ec68ffeb5a7b3d6623a1217cd3aa1>).

  - **Changes** - Added `YamlCommentCopier`, applied it to missing `settings.yml` keys in `ConfigsManager`, enabled comment parsing in `OraxenYaml` loaders, and added `YamlCommentCopierTest` coverage.

  

- **Added** shared internal command handling.

  - **Summary** - Added an Oraxen-owned `OraxenCommand` handler and migrated commands away from CommandAPI.

  - **Description** - Oraxen commands now route through an internal registration, permission, argument, execution, and tab-completion flow while preserving existing command behavior across supported server platforms and versions.

  - **Changes** - Added `OraxenCommand`, `CommandArguments`, and internal argument/suggestion classes, migrated all existing commands to the new handler, removed CommandAPI from Hopper, Gradle dependencies, run-server downloads, and plugin soft-depends, and unregisters commands on plugin disable.

- **Unified** Oraxen info commands.

  - **Summary** - Unified item, glyph, and block info lookups under `/oraxen info`.

  - **Description** - Oraxen now routes item, glyph, and block information through a shared info command handler instead of maintaining separate `/oraxen iteminfo`, `/oraxen glyphinfo`, and `/oraxen blockinfo` entry points.

  - **Changes** - Added `InfoCommand` with `/oraxen info <item|glyph|block> <id|all>`, preserved existing info output for all lookup types, updated command registration and help text, and removed the old individual info command classes.

    CleanShot 2026-06-05 at 05 08 02@2x

    CleanShot 2026-06-05 at 05 10 05@2x

    

- **Added** client version pack dispatch limits.

  - **Summary** - Added configurable client version filtering for resource pack dispatch.

  - **Description** - Oraxen can now skip sending the pack to players matching `Pack.dispatch.stop` rules using `<`, `=`, and `>` version checks, while still allowing eligible players to receive the correct pack, including in MultiVersionPacks mode.

  - **Changes** - Added `PackDispatchFilter`, wired filtering into Bukkit, MultiVersion, and pre-join pack dispatch, added the `Pack.dispatch.stop` setting, and added dispatch filter test coverage.

## Chores

- **Removed** stale NMS version mappings.

  - **Summary** - Removed obsolete 1.18 and 1.19 NMS mappings from `VersionUtil`.

  - **Description** - Oraxen no longer lists legacy 1.18.x and 1.19.x NMS versions in its supported NMS enum or version lookup map, avoiding misleading support detection for versions without active NMS modules.

  - **Changes** - Removed `v1_18_R1`, `v1_18_R2`, `v1_19_R1`, `v1_19_R2`, and `v1_19_R3` enum entries and deleted their corresponding `versionMap` registrations.

  

- **Cleaned up** replaced upload managers.

  - **Summary** - Unregistered old upload managers when replacing the active single-pack manager.

  - **Description** - `setUploadManager` now mirrors multi-version cleanup behavior by unregistering the previous manager before storing a different replacement, while preserving same-instance reuse without unregistering.

  - **Changes** - Updated `OraxenPlugin#setUploadManager`, removed manual unregistering at the mode-switch call site to avoid double-unregister side effects, added `OraxenPluginTest` coverage, and added the missing Adventure platform test dependency.

  

- **Renamed** the 26.1.2 NMS module.

  - **Summary** - Renamed the latest Paper NMS module from `v1_26_R1` to `v26_1_2`.

  - **Description** - Oraxen now uses MojangÃƒÂ¢Ã¢â€šÂ¬Ã¢â€žÂ¢s updated version naming for the 26.1.2 NMS handler while keeping NMS detection constrained to the exact supported release and preventing `1.21.11` from being treated as newer than `26.1.2`.

  - **Changes** - Renamed the Gradle module and Java package to `v26_1_2`, updated `VersionUtil` NMS mapping/detection, refreshed protocol/version labels to `26.1.2`, added version comparison/NMS detection tests, and made `JarReader` test-safe when the plugin is uninitialized.

- **Added** mechanic test coverage.

  - **Summary** - Added dedicated tests for mechanics under `core/src/test/java/io/th0rgal/oraxen/mechanics`.

  - **Description** - Oraxen now has expanded mechanic coverage across combat, cosmetic, farming, gameplay, and misc mechanics.

  - **Changes** - Added 40 separate mechanic test files, shared mechanic test support, a test Paper registry service loader, Spring Expression test dependency.

- **Cleaned up** runtime resource on disable.

  - **Summary** - Added explicit shutdown cleanup for upload managers, BukkitAudiences, HUD tasks, and resource pack generation state.

  - **Description** - `onDisable()` now clears runtime resources to avoid lingering listeners, tasks, upload state, audience instances, and in-progress pack generation during shutdown or reloads.

  - **Changes** - Added runtime cleanup in `OraxenPlugin`, closed `BukkitAudiences`, unregistered upload managers and HUD tasks, added resource pack shutdown handling, improved multi-version upload cancellation cleanup, and expanded `OraxenPluginTest` coverage.


pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/config/YamlCommentCopier.java
```diff
@@ -0,0 +1,185 @@
...
+                continue;
+            }
+            if (currentChar == '"' && !inSingleQuotes) {
+                inDoubleQuotes = !inDoubleQuotes;
+                continue;
+            }
+            if (currentChar == searchedChar && !inSingleQuotes && !inDoubleQuotes)
+                return index;
+        }
+        return -1;
```
miziusBot, line 155: <a href="#">P2</a> **Escaped single-quote not handled in `findUnquoted`**

In YAML, a single-quoted string escapes a literal apostrophe by doubling it: `key: 'it''s'`. The current scanner toggles `inSingleQuotes` on every `'`, so the second `''` would flip the flag back to `false` mid-string and cause the rest of the value to be parsed as unquoted. Any YAML key value (or inline comment anchor) that contains `''` would be misclassified, potentially yielding an incorrect separator index or a false positive hit on `:` or `#` inside the value.

core/src/main/java/io/th0rgal/oraxen/font/GlyphAppearance.java
```diff
@@ -18,6 +18,8 @@
  */
 public record GlyphAppearance(@NotNull Key font, @Nullable Integer shadowColor) {
 
+    private static boolean warnedShadowColorUnavailable;
```
miziusBot: <a href="#">P2</a> `warnedShadowColorUnavailable` is a non-`volatile` static field in a record. A Netty I/O thread calling `applyShadowColor` on a server that does not support `ShadowColor` will write `true` to this field, but another Netty thread may never observe the update due to a missing memory barrier. The practical consequence is multiple duplicate warnings instead of just one, but marking it `volatile` costs nothing and closes the visibility gap.

```suggestion
    private static volatile boolean warnedShadowColorUnavailable;
```

core/src/main/java/io/th0rgal/oraxen/pack/dispatch/PackDispatchFilter.java
```diff
@@ -0,0 +1,456 @@
...
+        for (Map.Entry<String, Integer> entry : PROTOCOL_BY_VERSION.entrySet()) {
+            ClientVersion version = ClientVersion.parse(entry.getKey()).orElse(null);
+            if (version == null) continue;
+            bestVersions.merge(entry.getValue(), version, (current, candidate) -> current.compareTo(candidate) <= 0 ? current : candidate);
+        }
+
+        TreeMap<Integer, ClientVersion> versions = new TreeMap<>();
+        versions.putAll(bestVersions);
+        return versions;
+    }
```
miziusBot, line 356: <a href="#">P2</a> **Reverse protocol map retains the _lowest_ version string per protocol**

The merge comparator `current.compareTo(candidate) <= 0 ? current : candidate` keeps the lexicographically smallest `ClientVersion`. For protocol 775 the candidates include both `26.1` and `1.26.1`; the comparison `[26,1]` vs `[1,26,1]` evaluates `26 > 1` so `1.26.1` wins. The debug log from `getVersionStringForProtocol` would therefore print `1.26.1` instead of the canonical `26.1.2` for connections on the latest Paper build. Functional version matching is unaffected (it uses `PROTOCOL_BY_VERSION` directly), but the misleading log message could confuse operators diagnosing dispatch filter behaviour.

core/src/main/resources/settings.yml
```diff
@@ -219,6 +219,7 @@ Pack:
     send_pre_join: false # Pre-join dispatch (Paper 1.21.7+). Takes priority over send_on_join when supported. Requires explicit opt-in.
     send_on_join: true # Join dispatch (after full join). Default mode for most servers.
     send_on_reload: true
+    stop: [] # Client versions that should not receive the pack. Supports '< 1.12', '= 1.21.11' and '> 1.21.11'.
```
miziusBot: <a href="#">P2</a> **`>` operator in `stop` list requires quoting in YAML**

The comment shows `'< 1.12'` and `'> 1.21.11'` with single quotes. If a user writes an unquoted entry such as `- > 1.21.11`, the YAML parser treats `>` in that block-sequence position as a folded block scalar indicator, yielding the string `1.21.11\n` rather than `> 1.21.11`. The rule pattern won't match and a warning will be logged. The same issue applies to `=` (which YAML can also misparse). Adding an explicit note that all three operators must be quoted would prevent silent misconfiguration.

core/src/main/java/io/th0rgal/oraxen/config/YamlCommentCopier.java
pxlarified: ✅ Fixed in 5ac943d11b58daedd620a890b9a94ab461e06e12.

core/src/main/java/io/th0rgal/oraxen/font/GlyphAppearance.java
pxlarified: ✅ Fixed in 5ac943d11b58daedd620a890b9a94ab461e06e12.

core/src/main/java/io/th0rgal/oraxen/pack/dispatch/PackDispatchFilter.java
pxlarified: ✅ Fixed in 5ac943d11b58daedd620a890b9a94ab461e06e12.

core/src/main/resources/settings.yml
pxlarified: ✅ Fixed in 5ac943d11b58daedd620a890b9a94ab461e06e12.

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockDurability.java
```diff
@@ -0,0 +1,104 @@
...
+        if (action == null || player.getGameMode() == GameMode.CREATIVE || ItemUtils.isInvalidItem(item)) return;
+
+        if (action.add() > 0) repairItem(player, item, action.add());
+        if (action.remove() > 0 && !ItemUtils.isInvalidItem(item)) damageItem(player, item, action.remove());
+
+        registerPendingVanillaDamageCancellation(player);
+    }
+
+    public static void setSuppressVanillaDamageCancellation(boolean suppress) {
+        if (suppress) SUPPRESS_VANILLA_DAMAGE_CANCELLATION.set(true);
```
miziusBot, line 38: <a href="#">P1</a> **Vanilla damage cancelled even when both `add` and `remove` are zero**

`registerPendingVanillaDamageCancellation` is called unconditionally whenever `action != null`. A user who writes `durability: {}` or `durability: { add: 0, remove: 0 }` will get a `DurabilityAction(0, 0)` from `parseDurabilityAction` (both `parseNonNegativeInt` calls default to 0), which is non-null. This silently suppresses the vanilla `PlayerItemDamageEvent` with no replacement damage, so the tool takes no durability loss when breaking the block at all.

The guard should check that at least one of the actions is meaningful before registering the cancellation.

```suggestion
        if (action.add() > 0) repairItem(player, item, action.add());
        if (action.remove() > 0 && !ItemUtils.isInvalidItem(item)) damageItem(player, item, action.remove());

        if (action.add() > 0 || action.remove() > 0) registerPendingVanillaDamageCancellation(player);
```

core/src/main/java/io/th0rgal/oraxen/font/GlyphAppearance.java
```diff
@@ -18,6 +18,8 @@
  */
 public record GlyphAppearance(@NotNull Key font, @Nullable Integer shadowColor) {
 
+    private static volatile boolean warnedShadowColorUnavailable;
+
```
miziusBot, line 23: <a href="#">P2</a> **Non-atomic check-then-set on `warnedShadowColorUnavailable`**

`volatile` guarantees visibility but not atomicity of the read-modify-write sequence. Two threads can both read `false`, both set the flag to `true`, and both emit the warning. In practice this is benign (duplicate log line), but standard practice for a one-shot flag is either `AtomicBoolean.compareAndSet` or a `synchronized` block around the check and set.

core/src/main/java/io/th0rgal/oraxen/utils/VersionUtil.java
```diff
@@ -118,11 +102,11 @@ public static NMSVersion getNMSVersion(MinecraftVersion version) {
...
 
-        // For 26.x series, use range-based matching to handle patch versions.
-        // Accept both canonical "26.x" and normalized "1.26.x" runtime shapes.
-        if ((version.getMajor() == 26 && version.getMinor() >= 1)
-                || (version.getMajor() == 1 && version.getMinor() == 26 && version.getBuild() >= 1)) {
-            return NMSVersion.v1_26_R1;
+        // Mojang switched the release version namespace after 1.21.11.
+        // Only bind this handler to the exact 26.1.2 release it was built against;
+        // future versions such as 26.2/26.2.1 should remain UNKNOWN until supported.
+        if (version.getMajor() == 26 && version.getMinor() == 1 && version.getBuild() == 2) {
```
miziusBot: <a href="#">P2</a> **Paper 26.1 and 26.1.1 now resolve to `UNKNOWN` NMS**

The previous `v1_26_R1` handler covered 26.1, 26.1.1, and 26.1.2 (all share protocol 775 and the same NMS structure per the old code comment). Restricting the new `v26_1_2` binding to the exact patch version `26.1.2` silently downgrades servers still on 26.1 or 26.1.1 to `NMSVersion.UNKNOWN`, disabling glyph injection, pre-join pack dispatch, and other NMS-backed features on those builds.

core/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/block/BlockDurability.java
pxlarified: ❌ This is intended behavior. Keep this as it is.

core/src/main/java/io/th0rgal/oraxen/font/GlyphAppearance.java
pxlarified: ✅ Fixed in e0ef3669.

core/src/main/java/io/th0rgal/oraxen/utils/VersionUtil.java
pxlarified: ✅ Fixed in e0ef3669.

PR #1839, opened 2026-06-16
v1.217.0

pxlarified:
## v1.217.0

* **Fixes**
  * **Validated** `/oraxen give` amounts (could crash when amount is `0` or negative).
  * **Fixed** unsupported block placement over Oraxen note blocks (shulker boxes, pistons and sticky pistons could restore custom note blocks to vanilla note blocks).
  * **Fixed** MiniMessage (`<red>`) parsing for custom painting titles and authors.
  * **Fixed** crafting recipes that use Oraxen backpack items as ingredients after the backpack has stored item data.
  * **Fixed** shaped Oraxen recipes in the 2x2 inventory crafting grid (prevented `ArrayIndexOutOfBoundsException` during prepare craft validation).
  * **Fixed** backpack-to-backpack crafting losing stored backpack contents.
  * **Fixed** BlockLocker furniture protection and furniture interaction access checks.
  * **Fixed** Folia world-mismatch errors in custom step and fall sound handling.
  * **Fixed** item tool component version handling (could crash on 1.20.5 or lower).
  * **Fixed** player head item definition fallback.
  * **Fixed** glyph shadow color handling.
  * **Fixed** unused block subtype activation.
  * **Guarded** the shaped block mechanic behind `1.20.4+`.
  * **Guarded** manager event cleanup when previous managers are not initialized.
  * **Fixed** internal command usage feedback to show the specific sub-command usage instead of the global sub-command list.
  * **Fixed** the stonecutting recipe builder opening a crafting table/workbench menu instead of a stonecutter menu.
  * **Fixed** custom paintings dropping the vanilla item when broken.
   * **Fixed** shaped block duplicate variation reloads (causing false warning messages).
   * **Fixed** legacy null texture atlas crashes (causing crashes on NeoForge clients).
   * **Fixed** Oraxen/BukkitAudiences message handling on Paper 26.2.

* **Features**
  * **Added** 26.2 support (pack format 88, 26.2 overlays/text shaders and safe 26.2 NMS item conversion).
  * **Added** 1.20.1 support.
  * **Migrated** to AntiGriefLib (replacing ProtectionLib), adding support for over 25 plugins, such as GriefDefender, GriefPrevention, NoBuildPlus and UltimateClaims.
  * **Added** a custom WorldGuard integration, so WorldGuard bypass permissions continue to allow protected-region placement and breaking.
  * **Added** support for copying bundled `settings.yml` comments when inserting newly added settings into existing configs.
  * **Added** shared internal Oraxen command handling and migrated commands away from CommandAPI.
  * **Added** configurable block break durability.

    ```yaml
    myblock:
      Mechanics:
        block:
          breaking:
            - else:
              hardness: 1
              durability:
                add: 0
                remove: 2
    ```
  * **Unified** item, glyph, and block info lookups under `/oraxen info`.

    605346532-e086646b-e860-4c36-8e3a-c9fc010854d8
  * **Added** painting hot reloading support and removed datapack generation (paintings are now handled via NMS instead of datapacks, Paper 1.21.5+).
  * **Added** configurable client version filtering for resource pack dispatch.

    ```yaml
    Pack:
      dispatch:
        stop: ["<1.18", "=26.1.2"] # These versions are excluded and won't receive a pack.
    ```

* **Chores**
  * **Removed** `.gitattributes`.
  * **Removed** stale NMS version mappings.
  * **Cleaned up** replaced upload managers.
  * **Cleaned up** runtime resources on disable.
  * **Migrated** away from per-version NMS handlers by consolidating modern NMS handling into the shared `nms` module and removing old per-version Paper/Spigot modules.
  * **Renamed** the latest Paper NMS module from `v1_26_R1` to `v26_1_2`.
  * **Removed** CommandAPI from Hopper, Gradle dependencies, run-server downloads and plugin soft-depends.
  * **Refactored** internal package structure for configs, glyphs, sounds, and paintings. 
  * **Cleaned up** obsolete `settings.yml` options.

* **Other**
  * **Added** mechanic test coverage (around 40 tests).
  * **Added** version loading test coverage (checks if the plugin loads fine on each Minecraft version).
  * **Added** settings removal handling to remove old settings options.
  * **Removed** support for Spigot servers.

* **Docs**
  * **Updated** `paintings.yml` docs for MiniMessage and Paper 1.21.5+ NMS painting support.
  * **Added** block breaking durability docs.
  * **Updated** the `README.md`.
  * **Removed** old GitHub issue templates, JitPack config, and generated schema file.
  * **Updated** contributors/README assets.


Th0rgal:
@bugbot review

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -1,47 +1,47 @@
...
+            BlockLockerMechanic blockLocker = getBlockLocker(block);
+            return blockLocker != null && blockLocker.canProtect() && blockLocker.getProtectionType() == type;
+        }
+    }
+
+    public static boolean canInteract(Player player, Block block) {
+        if (!CompatibilitiesManager.isCompatibilityEnabled("BlockLocker") || getBlockLocker(block) == null)
+            return true;
+
+        return BlockLockerAPIv2.isAllowed(player, block, true);
```
miziusBot, line 40: <a href="#">P1</a> **Note block and string block BlockLocker protection silently dropped**

The new `getBlockLocker` helper only queries `OraxenFurniture.getFurnitureMechanic(block)`, so any note block or string block that had a `block_locker` mechanic configured will no longer be recognized as protectable. Both `canProtect` overloads inside `BlockLockerProtection` call this helper, meaning existing BlockLocker signs placed on custom note/string blocks will stop being enforced after the upgrade and `canInteract` will always return `true` for those blocks. The previous implementation explicitly resolved `NoteBlockMechanic` and `StringBlockMechanic` and should be preserved alongside the new furniture path.

core/src/main/java/io/th0rgal/oraxen/pack/dispatch/PackDispatchFilter.java
```diff
@@ -0,0 +1,456 @@
...
+                Field field = current.getDeclaredField(fieldName);
+                field.setAccessible(true);
+                return field.get(object);
+            } catch (ReflectiveOperationException | RuntimeException ignored) {
+            }
+            current = current.getSuperclass();
+        }
+        return null;
+    }
+
```
miziusBot, line 281: <a href="#">P1</a> **MC 26.2 (protocol 776) absent from the dispatch protocol map**

`ProtocolVersion.java` adds `MC_26_2(776, 88, "26.2")`, but `createProtocolMap()` in `PackDispatchFilter` has no entry for version `"26.2"` or protocol `776`. When a 26.2 client connects, `findProtocolForVersion("26.2")` returns `null`, so every `VersionLimit` falls back to `clientVersion.compareTo(targetVersion)`. Because `VERSION_BY_PROTOCOL.floorEntry(776)` resolves to 775 → "26.1.2", even the `clientVersion` object used in the fallback path is wrong. A dispatch-stop rule like `"=26.2"` or `">26.1.2"` will therefore silently fail to exclude 26.2 clients.

Add `put(protocols, 776, "26.2");` (or the correct alias set) alongside the 775 entry to keep the filter table in sync with `ProtocolVersion`.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+    private static final NamespacedKey PLACED_PAINTING_ITEM = new NamespacedKey(OraxenPlugin.get(), "placed_painting_item");
+    private static final NamespacedKey PLACED_PAINTING_ITEM_ID = new NamespacedKey(OraxenPlugin.get(), "placed_painting_item_id");
+    private static final double DROP_MATCH_RADIUS_SQUARED = 4.0D;
+    private static final List<PendingPaintingDrop> PENDING_PAINTING_DROPS = new ArrayList<>();
+
+    @EventHandler(priority = EventPriority.HIGH, ignoreCancelled = true)
+    public void onPaintingInteract(PlayerInteractEvent event) {
+        if (event.getAction() != Action.RIGHT_CLICK_BLOCK) return;
+        if (event.useInteractedBlock() == Event.Result.DENY) return;
+        if (event.useItemInHand() == Event.Result.DENY) return;
```
miziusBot, line 59: <a href="#">P2</a> **`PENDING_PAINTING_DROPS` is a non-thread-safe `ArrayList` shared across event handlers and deferred tasks**

On Folia, `SchedulerUtil.runTaskLater` and entity events may execute on different region threads concurrently. `PENDING_PAINTING_DROPS` is a plain `ArrayList`, so concurrent `add`/`remove`/`removeIf`/`Iterator` access will produce `ConcurrentModificationException` or silent data loss. Replace it with `Collections.synchronizedList(new ArrayList<>())` or a `CopyOnWriteArrayList` (safe for the read-heavy usage pattern here).

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+            HangingPlaceEvent placeEvent = new HangingPlaceEvent(painting, player, clickedBlock, face, hand, item);
+            OraxenPlugin.get().getServer().getPluginManager().callEvent(placeEvent);
+            if (placeEvent.isCancelled()) return false;
+
+            storePaintingItem(painting, item);
+            consumeItem(player, hand);
+            placed = true;
+            return true;
+        } finally {
+            if (!placed) painting.remove();
```
miziusBot: <a href="#">P1</a> **Null painting reference after spawn**

`World.spawn()` returns `null` when the spawn event is cancelled by another plugin (e.g., a protection plugin, anti-cheat, or any listener that cancels `EntitySpawnEvent`). In that case `painting` is `null` on line 140, so `painting.setFacingDirection(face, true)` throws a `NullPointerException`. The `finally` block at line 153 also calls `painting.remove()` without a null guard, so both the happy path and the cleanup path will crash.

Add a null check immediately after the `spawn()` call and guard the `finally` block accordingly.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+    public void onPaintingItemSpawn(ItemSpawnEvent event) {
+        Item item = event.getEntity();
+        if (item.getItemStack().getType() != Material.PAINTING) return;
+
+        PendingPaintingDrop pendingDrop = findPendingDrop(item.getLocation());
+        if (pendingDrop == null) return;
+
+        replaceDrop(item, pendingDrop.itemStack());
+        PENDING_PAINTING_DROPS.remove(pendingDrop);
```
miziusBot, line 119: <a href="#">P2</a> **`PENDING_PAINTING_DROPS` is a plain `ArrayList` with concurrent access**

The static list is mutated by at least three event handlers (`onPaintingBreak` adds, `onPaintingDrop` removes via `removePendingDropNear`, `onPaintingItemSpawn` removes via `PENDING_PAINTING_DROPS.remove()`). On Folia or in any async context a concurrent read/write will throw `ConcurrentModificationException`. The scheduled removal via `SchedulerUtil.runTaskLater` could also race with the event-thread removals. Consider replacing the field with a `CopyOnWriteArrayList`, or synchronising access on a dedicated lock.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+        try {
+            if (!painting.setFacingDirection(face, true)) return false;
+            if (!painting.setArt(art, true)) return false;
+            if (!painting.isValid()) return false;
+
+            HangingPlaceEvent placeEvent = new HangingPlaceEvent(painting, player, clickedBlock, face, hand, item);
+            OraxenPlugin.get().getServer().getPluginManager().callEvent(placeEvent);
+            if (placeEvent.isCancelled()) return false;
+
+            storePaintingItem(painting, item);
```
miziusBot: <a href="#">P2</a> **`storePaintingItem` is called twice per successful `tryPlace`**

When `placeEvent` is fired at line 144, the `onPaintingPlace` listener (registered in the same class, `EventPriority.MONITOR`) runs and calls `storePaintingItem` for this very entity. Then `tryPlace` calls `storePaintingItem` directly at line 148. The second write overwrites the same values with the same data, so there is no functional breakage today, but it's a maintenance hazard. You can remove the direct call and rely on `onPaintingPlace` handling it, or remove the event-listener path for entities placed via `tryPlace`.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
pxlarified: ✅ This is an intended change.

core/src/main/java/io/th0rgal/oraxen/pack/dispatch/PackDispatchFilter.java
pxlarified: ✅ Fixed in 02780ba2.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
pxlarified: ✅ Fixed in c7bdaa54.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
pxlarified: ✅ Fixed in c7bdaa54.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
pxlarified: ✅ Fixed in c7bdaa54.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
pxlarified: ✅ Fixed in c7bdaa54.

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+
+        try {
+            Registry<Art> paintingRegistry = RegistryAccess.registryAccess()
+                    .getRegistry(RegistryKey.PAINTING_VARIANT);
+            Art art = paintingRegistry.get(variantKey);
+            if (art == null) Logs.logWarning("Unknown painting_variant '" + paintingVariant + "'");
+            return art;
+        } catch (Exception exception) {
+            Logs.logWarning("Failed to resolve painting_variant '" + paintingVariant + "'");
+            Logs.debug(exception);
```
miziusBot, line 175: <a href="#">P1</a> <a href="#">security</a> **Missing protection check before custom painting placement**

`tryPlace` never calls `AntiGriefLib.canBuild()`, so players can place custom paintings inside WorldGuard/GriefPrevention/etc. protected regions that would otherwise block them. Every other placement path in this PR (furniture, note blocks, string blocks, shaped blocks) explicitly checks `AntiGriefLib.canBuild(player, location)` before placing — this one is the only exception. Add the check before the entity is spawned, e.g., `if (!AntiGriefLib.canBuild(player, target.getLocation())) return false;`.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -1,47 +1,47 @@
...
+    }
+
+    private static BlockLockerMechanic getBlockLocker(Block block) {
+        FurnitureMechanic furnitureMechanic = OraxenFurniture.getFurnitureMechanic(block);
+        return furnitureMechanic != null ? furnitureMechanic.getBlockLocker() : null;
+    }
 }
```
miziusBot, line 47: <a href="#">P1</a> **BlockLocker support dropped for note blocks and string blocks**

The refactored `getBlockLocker` helper now only resolves a `BlockLockerMechanic` for furniture. The previous implementation also checked `NoteBlockMechanic` and `StringBlockMechanic`. Any block configured with a `blocklocker:` section under a note-block or string-block mechanic will silently lose BlockLocker protection after this change — `canProtect` will return `false` for those blocks and `BlockLockerProtection` will never register them as protectable.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+        pdc.set(PLACED_PAINTING_ITEM, DataType.ITEM_STACK, storedItem);
+
+        String itemId = OraxenItems.getIdByItem(storedItem);
+        if (itemId != null) pdc.set(PLACED_PAINTING_ITEM_ID, PersistentDataType.STRING, itemId);
+    }
+
+    private ItemStack getPlacedItem(HangingPlaceEvent event) {
+        ItemStack eventItem = getPlacedItemFromEvent(event);
+        if (!ItemUtils.isEmpty(eventItem)) return eventItem;
+
```
miziusBot, line 194: <a href="#">P2</a> **Painting is spawned before the HangingPlaceEvent is fired or permission is checked**

The entity is materialised in the world via `World.spawn()` (line ~164), then `setFacingDirection` / `setArt` are called a second time and `HangingPlaceEvent` is fired manually. While the `finally` block does call `painting.remove()` when placement fails, other plugins listening to `EntitySpawnEvent` or `CreatureSpawnEvent` will already have seen and reacted to the entity before it is removed. Consider deferring the spawn until after all validation passes, or using `World.spawn`'s pre-spawn consumer exclusively and removing the duplicate post-spawn calls to `setFacingDirection` and `setArt`.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
Th0rgal: ✅ Fixed in PR #1840 (commit e5449900b763d6cfac58070ac63f713c7de8146d)

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
Th0rgal: ✅ Fixed in PR #1840 (commit e5449900b763d6cfac58070ac63f713c7de8146d)

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
Th0rgal: ✅ Fixed in PR #1840 (commit e5449900b763d6cfac58070ac63f713c7de8146d)

PR #1840, opened 2026-06-22
Fix review findings for v1.217.0

Th0rgal:
### Bug 1: PackDispatchFilter protocol 776 (26.2) not fully mapped
Adds the missing 26.2 aliases (`26.2.0` and `1.26.2.0`) to the protocol 776 entry and covers them in `PackDispatchFilterTest`.

### Bug 2: BlockLockerCompatibility regression
Restores BlockLocker checks for Oraxen note blocks and string blocks before falling back to furniture. Also restores the BlockLocker mechanic fields/getters on `NoteBlockMechanic` and `StringBlockMechanic`, matching the previous API shape used by compatibility code.

### Bug 3: Custom painting tag registration
Updates custom painting registry reflection to resolve the newer `IdMap.register(IdMap, Identifier, Object)` mapping when the older `Registry.register(Registry, ResourceLocation, Object)` signature is unavailable.

## Verification
- `./gradlew clean build -x test`
- `./gradlew :core:test --tests io.th0rgal.oraxen.pack.dispatch.PackDispatchFilterTest`

---

> [!NOTE]
> **High Risk**
> Replacing CommandAPI and collapsing NMS modules touches every command and version-specific code path; custom painting and pack-dispatch logic rely on fragile registry/reflection behavior across Minecraft versions.
> 
> **Overview**
> This is a **major release-style refactor** plus three **targeted fixes** called out for review of #1839.
> 
> **Architecture & dependencies:** Gradle now depends on one **`:nms`** module instead of many version-specific NMS subprojects; **CommandAPI** is removed from the build, run server, and soft-dependencies. **ProtectionLib** is replaced by in-repo **AntiGriefLib** (with Momirealms repo). Config-related packages move under **`configs`**, **`fonts`**, and **`sounds`**.
> 
> **Commands:** All `/oraxen` subcommands are wired through a new **`OraxenCommand`** stack (routing, permissions, tab completion, unregister on disable). Separate **`iteminfo` / `glyphinfo` / `blockinfo`** commands are merged into **`/oraxen info`**. **`give`** validates amount and caps inventory fill; help text is updated.
> 
> **Gameplay & lifecycle:** **Custom paintings** load from config, register at runtime via NMS reflection (with **`IdMap.register`** fallback for newer APIs), and support **`/oraxen reload paintings`**. **`Pack.dispatch.stop`** filtering skips pack send for blocked client versions; protocol **776** now includes **`26.2.0`** / **`1.26.2.0`** aliases (with tests). **BlockLocker** again resolves note/string block mechanics before furniture and exposes **`canInteract`**. Block breaks apply configured **tool durability**. Enable/disable paths add supported-server checks, painting listener, and cleaner upload/font/hud teardown.
> 
> **Docs/repo hygiene:** README is rewritten (features, examples, wiki links); GitHub issue templates and some contributor onboarding text are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77358816f2b7f025713ac198a44d141ae213d717. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Th0rgal:
@bugbot review

Th0rgal:
@mizius-bot review

Th0rgal:
@mizius-bot review

Th0rgal:
@bugbot review

Th0rgal:
@bugbot review

Th0rgal:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingListener.java
```diff
@@ -0,0 +1,334 @@
...
+        boolean placed = false;
+        try {
+            if (!painting.setFacingDirection(face, true)) return false;
+            if (!painting.setArt(art, true)) return false;
+            if (!painting.isValid()) return false;
+
+            HangingPlaceEvent placeEvent = new HangingPlaceEvent(painting, player, clickedBlock, face, hand, item);
+            OraxenPlugin.get().getServer().getPluginManager().callEvent(placeEvent);
+            if (placeEvent.isCancelled()) return false;
+
```
pxlarified, line 147: <a href="#">P1</a> <a href="#">security</a> **Missing protection check before custom painting placement**

`tryPlace` never calls `AntiGriefLib.canBuild()`, so players can place custom paintings inside WorldGuard/GriefPrevention/etc. protected regions that would otherwise block them. Every other placement path in this PR (furniture, note blocks, string blocks, shaped blocks) explicitly checks `AntiGriefLib.canBuild(player, location)` before placing — this one is the only exception. Add the check before the entity is spawned, e.g., `if (!AntiGriefLib.canBuild(player, target.getLocation())) return false;`.

pxlarified, line 147: <a href="#">P2</a> **Painting is spawned before the HangingPlaceEvent is fired or permission is checked**

The entity is materialised in the world via `World.spawn()` (line ~164), then `setFacingDirection` / `setArt` are called a second time and `HangingPlaceEvent` is fired manually. While the `finally` block does call `painting.remove()` when placement fails, other plugins listening to `EntitySpawnEvent` or `CreatureSpawnEvent` will already have seen and reacted to the entity before it is removed. Consider deferring the spawn until after all validation passes, or using `World.spawn`'s pre-spawn consumer exclusively and removing the duplicate post-spawn calls to `setFacingDirection` and `setArt`.

pxlarified:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -11,37 +12,45 @@
...
+            return blockLocker != null && blockLocker.canProtect() && blockLocker.getProtectionType() == type;
+        }
+    }
+
+    public static boolean canInteract(Player player, Block block) {
+        if (!CompatibilitiesManager.isCompatibilityEnabled("BlockLocker") || getBlockLocker(block) == null)
+            return true;
+
+        return BlockLockerAPIv2.isAllowed(player, block, true);
+    }
```
miziusBot, line 54: <a href="#">P2</a> **`canInteract` skips `canProtect()` guard**

`getBlockLocker(block)` returns a non-null `BlockLockerMechanic` whenever the block has any `blocklocker` config section, even when that section has `can_protect: false`. The method then unconditionally calls `BlockLockerAPIv2.isAllowed`, which is inconsistent with `BlockLockerProtection.canProtect(Block)` (lines 28–29), which correctly gates on both null-check and `canProtect()`. In practice, BlockLocker won't have a sign-protection entry for an opt-out block so `isAllowed` returns `true` and the caller is unblocked — but a block with `can_protect: false` should short-circuit here the same way it does in `canProtect(Block)`, without touching the BlockLocker API at all.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -11,37 +12,45 @@
...
-                StringBlockMechanic stringMechanic = OraxenBlocks.getStringMechanic(block);
-                if (stringMechanic != null) blockLocker = stringMechanic.getBlockLocker();
-                FurnitureMechanic furnitureMechanic = OraxenFurniture.getFurnitureMechanic(block);
-                if (furnitureMechanic != null) blockLocker = furnitureMechanic.getBlockLocker();
-
-                return blockLocker != null && blockLocker.canProtect() && blockLocker.getProtectionType() == type;
-            }});
+        BlockLockerAPIv2.getPlugin().getChestSettings().getExtraProtectables()
+                .removeIf(BlockLockerProtection.class::isInstance);
+        BlockLockerAPIv2.getPlugin().getChestSettings().getExtraProtectables().add(new BlockLockerProtection());
```
miziusBot, line 21: <a href="#">P2</a> **`removeIf` won't clean up across classloader boundaries**

`BlockLockerProtection.class::isInstance` resolves the `Class` object from the current plugin classloader. When Oraxen is hot-reloaded (e.g., via PlugMan or any mechanism that discards and recreates the plugin classloader), the old `BlockLockerProtection` instances were loaded by a previous classloader whose `Class` object is a different identity from the new `BlockLockerProtection.class`. `isInstance` therefore returns `false` for them, the `removeIf` silently no-ops, and a duplicate handler accumulates in BlockLocker's extra-protectables list on every reload cycle.

A classloader-safe guard would compare by class name rather than identity: `p -> p.getClass().getName().equals(BlockLockerProtection.class.getName())`.

This does not affect normal `/oraxen reload` (same classloader); it only matters for full plugin hot-reloads.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,554 @@
...
+                if (oldVariant != null) byValue.remove(oldVariant);
+                byValue.put(newVariant, holder);
+
+                Object toIdObject = toIdField.get(registry);
+                Method removeInt = toIdObject.getClass().getMethod("removeInt", Object.class);
+                Method put = toIdObject.getClass().getMethod("put", Object.class, int.class);
+                if (id >= 0 && oldVariant != null) {
+                    removeInt.invoke(toIdObject, oldVariant);
+                    put.invoke(toIdObject, newVariant, id);
+                }
```
miziusBot: <a href="#">P2</a> **`removeInt`/`put` method handles looked up per painting variant**

`toIdObject.getClass().getMethod("removeInt", ...)` and `getMethod("put", ...)` are resolved reflectively on every call to `replaceVariant`, once per painting per reload. Since these are the same NMS methods every time, they can be cached alongside the other `Method` fields in `Ref` to avoid repeated lookups. This is a minor inefficiency only observable if hundreds of paintings are reloaded frequently, but is worth noting for consistency with the rest of the `Ref` field pattern.

Th0rgal:
@mizius-bot review

Th0rgal:
@bugbot review

Th0rgal:
@bugbot review

Th0rgal:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,624 @@
...
+
+        private static Method optionalMethod(Class<?> owner, String name, Class<?>... parameterTypes) {
+            try {
+                return owner.getMethod(name, parameterTypes);
+            } catch (NoSuchMethodException ignored) {
+                return null;
+            }
+        }
+
+        private Method registryRegisterMethod() throws NoSuchMethodException {
```
miziusBot: <a href="#">P1</a> **Verify `IdMap.register` is a static method**

The fallback resolves `idMapClass.getMethod("register", idMapClass, resourceLocationClass, Object.class)` and later calls it as `registryRegisterMethod.invoke(null, registry, location, variant)` — the `null` receiver is valid only for static methods. If `IdMap.register(IdMap, Identifier, Object)` is actually an instance method on `IdMap` in the target Paper build, the invocation would throw `IllegalArgumentException: object is not an instance of declaring class` at runtime for every painting injection. Worth a quick check against the NMS source of the Paper version this targets before shipping.

Th0rgal:
@cursor review

Th0rgal:
@mizius-bot review

Th0rgal:
@bugbot review

Th0rgal:
bugbot run

Th0rgal:
@mizius-bot review

Th0rgal:
@mizius-bot review

pxlarified:
Note that the mizius bot takes a bit longer to review, around 10-15m per review.

Th0rgal:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -11,37 +12,55 @@
...
+    private static BlockLockerMechanic getBlockLocker(Block block) {
+        NoteBlockMechanic noteBlockMechanic = OraxenBlocks.getNoteBlockMechanic(block);
+        if (noteBlockMechanic != null) return noteBlockMechanic.getBlockLocker();
+
+        StringBlockMechanic stringBlockMechanic = OraxenBlocks.getStringMechanic(block);
+        if (stringBlockMechanic != null) return stringBlockMechanic.getBlockLocker();
+
+        FurnitureMechanic furnitureMechanic = OraxenFurniture.getFurnitureMechanic(block);
+        return furnitureMechanic != null ? furnitureMechanic.getBlockLocker() : null;
+    }
```
miziusBot, line 72: <a href="#">P2</a> **Early return masks null blockLocker on the first mechanic found**

When a `NoteBlockMechanic` or `StringBlockMechanic` is detected at the block but has no BlockLocker sub-mechanic configured (i.e., `getBlockLocker()` returns `null`), the method now returns `null` immediately without checking the remaining mechanic types. In the original code all three lookups always ran, so the last non-null blockLocker would win.

In practice a single block position can only host one mechanic type at a time, so the early return is equivalent today. However, if future mechanics allow coexistence (e.g., furniture sharing a note-block position), a note block without a BlockLocker config would silently shadow any furniture BlockLocker. Consider guarding the early return only when `getBlockLocker()` is non-null to keep the intent explicit.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -11,37 +12,55 @@
...
+
+    public static boolean canInteract(Player player, Block block) {
+        return canInteract(player, block, getBlockLocker(block));
+    }
+
+    public static boolean canInteract(Player player, Block block, FurnitureMechanic furnitureMechanic) {
+        BlockLockerMechanic blockLocker = furnitureMechanic != null ? furnitureMechanic.getBlockLocker() : null;
+        if (blockLocker == null) blockLocker = getBlockLocker(block);
+        return canInteract(player, block, blockLocker);
```
miziusBot, line 46: <a href="#">P2</a> **Redundant furniture lookup when explicit mechanic has no BlockLocker**

When `furnitureMechanic` is non-null but its `getBlockLocker()` returns `null`, the fallback `getBlockLocker(block)` re-queries `OraxenFurniture.getFurnitureMechanic(block)` and will find the same mechanic, receiving the same `null` result. More importantly, if the caller explicitly passes a `FurnitureMechanic` that has no BlockLocker, the fallback could in theory return a BlockLocker from an unrelated mechanic at the same position (e.g., a note block), producing an unexpected denial. Restricting the fallback to note/string block checks only when a furniture mechanic was explicitly provided would make the intent unambiguous.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,645 @@
...
+            try {
+                return RegistryRegisterMethod.of(registryClass.getMethod("register", registryClass, resourceLocationClass, Object.class));
+            } catch (NoSuchMethodException ignored) {
+                try {
+                    return RegistryRegisterMethod.of(registryClass.getMethod("register", idMapClass, resourceLocationClass, Object.class));
+                } catch (NoSuchMethodException ignoredAgain) {
+                    return RegistryRegisterMethod.of(idMapClass.getMethod("register", resourceLocationClass, Object.class));
+                }
+            }
+        }
```
miziusBot, line 416: <a href="#">P2</a> **Third fallback misroutes if `IdMap.register(ResourceLocation, Object)` is a static method**

`RegistryRegisterMethod.of` sets `passesRegistry = (paramCount == 3)` and `staticMethod = Modifier.isStatic(...)`. For the third fallback (`IdMap.register(ResourceLocation, Object)`, 2 params), `passesRegistry` is `false`. If this method happens to be a static interface method (possible with Java 8+ default/static interface methods in older remapped NMS builds), the resulting invoke path is `method.invoke(null, location, variant)` — the registry is never passed, so registration would silently fail or throw. This is a narrow edge case (NMS `IdMap.register` is almost certainly an instance method), but a defensive guard or comment would prevent silent failure on unexpected API shapes.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
miziusBot, line 72: <a href="#">P2</a> **Early return masks null blockLocker on the first mechanic found**

When a `NoteBlockMechanic` or `StringBlockMechanic` is detected at the block but has no BlockLocker sub-mechanic configured (i.e., `getBlockLocker()` returns `null`), the method now returns `null` immediately without checking the remaining mechanic types. In the original code all three lookups always ran, so the last non-null blockLocker would win.

In practice a single block position can only host one mechanic type at a time, so the early return is equivalent today. However, if future mechanics allow coexistence (e.g., furniture sharing a note-block position), a note block without a BlockLocker config would silently shadow any furniture BlockLocker. Consider guarding the early return only when `getBlockLocker()` is non-null to keep the intent explicit.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
miziusBot, line 46: <a href="#">P2</a> **Redundant furniture lookup when explicit mechanic has no BlockLocker**

When `furnitureMechanic` is non-null but its `getBlockLocker()` returns `null`, the fallback `getBlockLocker(block)` re-queries `OraxenFurniture.getFurnitureMechanic(block)` and will find the same mechanic, receiving the same `null` result. More importantly, if the caller explicitly passes a `FurnitureMechanic` that has no BlockLocker, the fallback could in theory return a BlockLocker from an unrelated mechanic at the same position (e.g., a note block), producing an unexpected denial. Restricting the fallback to note/string block checks only when a furniture mechanic was explicitly provided would make the intent unambiguous.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
miziusBot, line 416: <a href="#">P2</a> **Third fallback misroutes if `IdMap.register(ResourceLocation, Object)` is a static method**

`RegistryRegisterMethod.of` sets `passesRegistry = (paramCount == 3)` and `staticMethod = Modifier.isStatic(...)`. For the third fallback (`IdMap.register(ResourceLocation, Object)`, 2 params), `passesRegistry` is `false`. If this method happens to be a static interface method (possible with Java 8+ default/static interface methods in older remapped NMS builds), the resulting invoke path is `method.invoke(null, location, variant)` — the registry is never passed, so registration would silently fail or throw. This is a narrow edge case (NMS `IdMap.register` is almost certainly an instance method), but a defensive guard or comment would prevent silent failure on unexpected API shapes.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
miziusBot, line 72: <a href="#">P2</a> **Early return masks null blockLocker on the first mechanic found**

When a `NoteBlockMechanic` or `StringBlockMechanic` is detected at the block but has no BlockLocker sub-mechanic configured (i.e., `getBlockLocker()` returns `null`), the method now returns `null` immediately without checking the remaining mechanic types. In the original code all three lookups always ran, so the last non-null blockLocker would win.

In practice a single block position can only host one mechanic type at a time, so the early return is equivalent today. However, if future mechanics allow coexistence (e.g., furniture sharing a note-block position), a note block without a BlockLocker config would silently shadow any furniture BlockLocker. Consider guarding the early return only when `getBlockLocker()` is non-null to keep the intent explicit.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
miziusBot, line 46: <a href="#">P2</a> **Redundant furniture lookup when explicit mechanic has no BlockLocker**

When `furnitureMechanic` is non-null but its `getBlockLocker()` returns `null`, the fallback `getBlockLocker(block)` re-queries `OraxenFurniture.getFurnitureMechanic(block)` and will find the same mechanic, receiving the same `null` result. More importantly, if the caller explicitly passes a `FurnitureMechanic` that has no BlockLocker, the fallback could in theory return a BlockLocker from an unrelated mechanic at the same position (e.g., a note block), producing an unexpected denial. Restricting the fallback to note/string block checks only when a furniture mechanic was explicitly provided would make the intent unambiguous.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
miziusBot, line 416: <a href="#">P2</a> **Third fallback misroutes if `IdMap.register(ResourceLocation, Object)` is a static method**

`RegistryRegisterMethod.of` sets `passesRegistry = (paramCount == 3)` and `staticMethod = Modifier.isStatic(...)`. For the third fallback (`IdMap.register(ResourceLocation, Object)`, 2 params), `passesRegistry` is `false`. If this method happens to be a static interface method (possible with Java 8+ default/static interface methods in older remapped NMS builds), the resulting invoke path is `method.invoke(null, location, variant)` — the registry is never passed, so registration would silently fail or throw. This is a narrow edge case (NMS `IdMap.register` is almost certainly an instance method), but a defensive guard or comment would prevent silent failure on unexpected API shapes.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
miziusBot, line 72: <a href="#">P2</a> **Early return masks null blockLocker on the first mechanic found**

When a `NoteBlockMechanic` or `StringBlockMechanic` is detected at the block but has no BlockLocker sub-mechanic configured (i.e., `getBlockLocker()` returns `null`), the method now returns `null` immediately without checking the remaining mechanic types. In the original code all three lookups always ran, so the last non-null blockLocker would win.

In practice a single block position can only host one mechanic type at a time, so the early return is equivalent today. However, if future mechanics allow coexistence (e.g., furniture sharing a note-block position), a note block without a BlockLocker config would silently shadow any furniture BlockLocker. Consider guarding the early return only when `getBlockLocker()` is non-null to keep the intent explicit.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
miziusBot, line 46: <a href="#">P2</a> **Redundant furniture lookup when explicit mechanic has no BlockLocker**

When `furnitureMechanic` is non-null but its `getBlockLocker()` returns `null`, the fallback `getBlockLocker(block)` re-queries `OraxenFurniture.getFurnitureMechanic(block)` and will find the same mechanic, receiving the same `null` result. More importantly, if the caller explicitly passes a `FurnitureMechanic` that has no BlockLocker, the fallback could in theory return a BlockLocker from an unrelated mechanic at the same position (e.g., a note block), producing an unexpected denial. Restricting the fallback to note/string block checks only when a furniture mechanic was explicitly provided would make the intent unambiguous.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
miziusBot, line 416: <a href="#">P2</a> **Third fallback misroutes if `IdMap.register(ResourceLocation, Object)` is a static method**

`RegistryRegisterMethod.of` sets `passesRegistry = (paramCount == 3)` and `staticMethod = Modifier.isStatic(...)`. For the third fallback (`IdMap.register(ResourceLocation, Object)`, 2 params), `passesRegistry` is `false`. If this method happens to be a static interface method (possible with Java 8+ default/static interface methods in older remapped NMS builds), the resulting invoke path is `method.invoke(null, location, variant)` — the registry is never passed, so registration would silently fail or throw. This is a narrow edge case (NMS `IdMap.register` is almost certainly an instance method), but a defensive guard or comment would prevent silent failure on unexpected API shapes.

pxlarified:
Yeah that happens if the review command gets spammed xD

Th0rgal:
@mizius-bot review

Th0rgal:
bugbot run

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,647 @@
...
+                int sep = id.indexOf(':');
+                return resourceLocationClass.getMethod("fromNamespaceAndPath", String.class, String.class)
+                        .invoke(null, id.substring(0, sep), id.substring(sep + 1));
+            }
+        }
+
+        private Object resourceKey(Object location) throws ReflectiveOperationException {
+            return resourceKeyCreateMethod.invoke(null, paintingVariantRegistryKey, location);
+        }
+
```
miziusBot, line 469: <a href="#">P2</a> **Missing colon guard in `fromNamespaceAndPath` fallback**

`resourceLocation(String id)` falls back to `fromNamespaceAndPath` by splitting on the first `:`. If a variant ID has no `:`, `id.indexOf(':')` returns `-1` and `id.substring(0, -1)` throws `StringIndexOutOfBoundsException` before the reflective call is even made. That exception is a `RuntimeException`, so it is caught by the outer reload handler and silently swallowed — but the painting is silently dropped instead of being registered. In practice all painting keys should be namespaced, but a misconfigured or manually-added `paintings.yml` entry that omits the namespace would fail here without a clear log message.

core/src/main/java/io/th0rgal/oraxen/compatibilities/provided/blocklocker/BlockLockerCompatibility.java
```diff
@@ -11,37 +12,62 @@
...
+
+    public static boolean canInteract(Player player, Block block) {
+        return canInteract(player, block, getBlockLocker(block));
+    }
+
+    public static boolean canInteract(Player player, Block block, FurnitureMechanic furnitureMechanic) {
+        BlockLockerMechanic blockLocker = furnitureMechanic != null ? furnitureMechanic.getBlockLocker() : null;
+        if (blockLocker == null) blockLocker = getBlockLocker(block, false);
+        return canInteract(player, block, blockLocker);
+    }
```
miziusBot, line 47: <a href="#">P2</a> **Null furniture mechanic silently skips furniture protection**

`canInteract(Player, Block, FurnitureMechanic)` calls `getBlockLocker(block, false)` when `furnitureMechanic` is `null` (or its `getBlockLocker()` returns `null`), which explicitly excludes furniture blocks. Any call site that passes `null` because it has not yet resolved the mechanic — or where the lookup returned `null` for an unexpected reason — will bypass BlockLocker protection for furniture blocks at that location. The single-argument `canInteract(Player, Block)` correctly passes `includeFurniture = true`, so callers that go through that path are safe; it is only this two-argument overload where the silent skip can occur.

Th0rgal:
@mizius-bot review

Th0rgal:
bugbot run

Th0rgal:
bugbot run

Th0rgal:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,647 @@
...
+            }
+        }
+
+        private record RegistryRegisterMethod(Method method, boolean staticMethod, boolean passesRegistry) {
+
+            private static RegistryRegisterMethod of(Method method) {
+                boolean staticMethod = Modifier.isStatic(method.getModifiers());
+                return new RegistryRegisterMethod(method, staticMethod, method.getParameterCount() == 3);
+            }
+
```
miziusBot, line 372: <a href="#">P2</a> **Static 2-param guard re-throws into `Ref` constructor**

`RegistryRegisterMethod.of` throws `NoSuchMethodException` when it finds a static method with fewer than 3 parameters. If a future NMS build exposes `IdMap.register(ResourceLocation, Object)` as a static interface method rather than an instance method, the third fallback in `registryRegisterMethod()` would hit this guard, the exception would propagate through the `Ref()` constructor uncaught, and `Ref.get()` would rethrow — ultimately silencing all custom paintings with only a logged warning. The degradation is graceful (no crash), but the constraint is invisible at the call site. A comment explaining that this guard assumes the 2-param fallback is always an instance method would make the intent clearer and help future maintainers spot the assumption.

pxlarified:
Note that the other review is still processing.

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,647 @@
...
+                        method.invoke(null, location, variant);
+                    }
+                    return;
+                }
+
+                if (passesRegistry) {
+                    method.invoke(registry, registry, location, variant);
+                } else {
+                    method.invoke(registry, location, variant);
+                }
```
miziusBot, line 394: <a href="#">P2</a> **Non-static `passesRegistry` branch passes registry twice**

The instance-method branch at line 360 calls `method.invoke(registry, registry, location, variant)`, forwarding `registry` as both the receiver instance and the first argument. This would only be reached if a 3-parameter `register` method is resolved as non-static — all known NMS `register` helpers (the two lookups above this fallback in `registryRegisterMethod()`) are in fact `static`, so this branch is unreachable in practice. If a future NMS remap ever produces a non-static 3-param overload, the invocation would silently pass the wrong argument in position 0. Consider asserting `staticMethod` is true after `of(...)`, or throwing a `NoSuchMethodException` if the resolved method is unexpectedly non-static.

Th0rgal:
bugbot run

Th0rgal:
@mizius-bot review

core/src/main/java/io/th0rgal/oraxen/paintings/CustomPaintingRegistry.java
```diff
@@ -0,0 +1,646 @@
...
+        private RegistryRegisterMethod registryRegisterMethod() throws NoSuchMethodException {
+            try {
+                return RegistryRegisterMethod.of(registryClass.getMethod("register", registryClass, resourceLocationClass, Object.class));
+            } catch (NoSuchMethodException ignored) {
+                try {
+                    return RegistryRegisterMethod.of(registryClass.getMethod("register", idMapClass, resourceLocationClass, Object.class));
+                } catch (NoSuchMethodException ignoredAgain) {
+                    return RegistryRegisterMethod.of(idMapClass.getMethod("register", resourceLocationClass, Object.class));
+                }
+            }
```
miziusBot, line 415: <a href="#">P1</a> **Fallback lookup uses wrong class for the second-tier register method**

The PR description says the new intermediate fallback resolves `IdMap.register(IdMap, Identifier, Object)` — a static method on `IdMap`. But the code calls `registryClass.getMethod("register", idMapClass, ...)`, which searches `Registry`, not `IdMap`. If in the target NMS version the method is truly a static on `IdMap` (not on `Registry`), `registryClass.getMethod` will throw `NoSuchMethodException`, skip this tier, and fall directly to the instance-method lookup `idMapClass.getMethod("register", resourceLocationClass, Object.class)`. If that instance method is also absent on the newer mapping, `Ref` construction throws, `reload()` catches it, logs a warning, and all custom paintings silently stop loading.

Please confirm whether the method is declared on `Registry` (with an `IdMap` parameter type) or on `IdMap` itself. If the latter, the second fallback should use `idMapClass.getMethod("register", idMapClass, resourceLocationClass, Object.class)`.

Th0rgal:
bugbot run

Th0rgal:
@mizius-bot review

PR #1847, opened 2026-07-28
fix: account for furniture rotation in seat yaw

OlivierChiasson:
Fix furniture seat yaw rotation for rotated furniture

## Problem
When using furniture seats with an explicit `yaw` value, the seated character's facing direction was in furniture-local yaw but was being treated as world yaw. This meant seats only faced correctly when the furniture was at yaw 0.

For example, a picnic table with seats configured with yaw `90` and `270` to face inward would work at yaw 0, but break at yaw 90/180/270.

## Root Cause
`FurnitureSeat.yaw(float)` returned the configured seat yaw directly without transforming it from furniture-local space to world space. The seat's local facing direction was not being rotated by the furniture's world yaw.

## Changes
- `FurnitureMechanic.java`: Added `FurnitureSeat.rotatedYaw(float furnitureYaw)` which transforms the configured seat yaw from furniture-local space to world space.
- Updated `spawnSeat` and `updateFurnitureSeats` to use `rotatedYaw` instead of `yaw` when setting the seat's world rotation.
- `FurnitureSeatTest.java`: Added `seatYawRotatesWithFurniture` parameterized test covering all four cardinal furniture yaws.

## Result
Seat yaw is now correctly rotated with the furniture. A seat configured with `yaw: 90` faces 90° relative to the furniture's forward direction, which is properly transformed to the correct world-space yaw based on the furniture's placement rotation.

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/furniture/FurnitureMechanic.java
+++ b/src/main/java/io/th0rgal/oraxen/mechanics/provided/gameplay/furniture/FurnitureMechanic.java
@@ -134,6 +134,10 @@ public class FurnitureMechanic extends Mechanic {
         public float yaw(float fallbackYaw) {
             return yaw != null ? yaw : fallbackYaw;
         }
+
+        public float rotatedYaw(float furnitureYaw) {
+            return yaw != null ? ((yaw + furnitureYaw) % 360) : furnitureYaw;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


pxlarified:
Hi, thanks a lot for the contribution. I really appreciate it. I cherry-picked it into the `1.219.0` branch, where it will be shipped in the release. Feel free to close this pull request.

PR #1852, opened 2026-08-25
feat: add backpack cosmetics support for armor stand chest display

OlivierChiasson:
image

## Summary
Adds support for placing backpack cosmetic items on armor stands' chest slots. Previously, equipping a backpack cosmetic on an armor stand did nothing (item would be placed on armor stand but nothing rendered). Now, when such an item is placed on an armor stand's chest, a packet-based invisible companion object is spawned and mounted on the armor stand, displaying the backpack model. This completes the armor stack display use case (head + chest cosmetic on a single stand).

## Problem
Backpack-cosmetic items placed in an armor stand's chest slot were ignored because the existing implementation only handled display against a player target (BackpackCosmeticManager), and only propped on the player's own body. There was no mechanism for showing the chest piece as a floating companion object on an armor stand.

## Root Cause
The armor-stand path is separate from the player path (which emits the backpack as a companion object with a dedicated model and mount behavior). The listener had no handling for the armor-stand chest slot case.

## Changes
- `BackpackCosmeticListener.java`: Added detection of backpack-cosmetic items in an armor stand's chest slot; when present, spawn a packet-based companion object (with the item's model and the small flag from the mechanic) and mount it to the armor stand. The display follows the stand’s position and rotation automatically via the mount, and the initial head rotation is set to match the stand to avoid the body-vs-head offset (the same fix used for the player path).
- Added a periodic refresh and per-viewer tracking so the display reliably restores for players who join or re-enter range after the stand is already built, using the same pattern as the player path's viewer management.
- Added a short deferral before spawning to avoid a race where the stand's position is read before it is fully loaded, which could cause the display to spawn offset.
- Added standard cleanup on stand death/removal and when the chest item is changed or removed.
## Result
Full body armor display is now possible on a single armor stand: equipping a helmet-style cosmetic as the head and a backpack cosmetic as the chest displays both. The chest component correctly follows the stand and persists for viewers across joins, relocations, and reloads.

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/mechanics/provided/cosmetic/backpack/BackpackCosmeticFactory.java
+++ b/src/main/java/io/th0rgal/oraxen/mechanics/provided/cosmetic/backpack/BackpackCosmeticFactory.java
@@ -64,6 +64,11 @@ public class BackpackCosmeticFactory extends MechanicFactory {
         SchedulerUtil.ScheduledTask refreshTask = SchedulerUtil.runTaskTimer(20L, 20L, manager::refreshAllViewers);
         MechanicsManager.registerTask(getMechanicID(), refreshTask);
 
+        // Armor stand display refresh (every 20 ticks = 1 second) for join/view-range
+        // restoration of backpack cosmetics placed in armor stand chest slots.
+        SchedulerUtil.ScheduledTask armorStandRefreshTask = SchedulerUtil.runTaskTimer(20L, 20L, listener::refreshArmorStandDisplays);
+        MechanicsManager.registerTask(getMechanicID(), armorStandRefreshTask);
+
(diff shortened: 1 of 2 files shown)


pxlarified:
There are a few issues I’ll resolve first. It targets `1.218.0`, but conflicts with both `1.219.0` and `1.220.0`, and I don’t believe it is currently Folia-safe.

It also doesn’t preserve existing armor-stand passengers when sending mount packets. This is important because “sit” plugins, which allow players to stack on top of each other, would otherwise cause the backpack to detach when someone stacks onto another player.

That said, this is a great idea and I really like it. I’ll add it on top of `1.220.0` while preserving your commits.

pxlarified:
I think you can close this btw., I already added it to `1.220.0`. Again, thank you so much for your contribution.

PR #1853, opened 2026-08-26
fix: restore creative middle-click pick of Oraxen furniture

OlivierChiasson:
## Summary
Restores middle-click (pick block) on an Oraxen furniture in creative mode giving the actual
Oraxen furniture item instead of a plain `BARRIER`. Since MC 1.21.4, creative pick-block no
longer fires `InventoryCreativeEvent`, which silently broke the existing middle-click handling
in `FurnitureListener`. This fix intercepts the new pick-item packets and substitutes the real
furniture item.

## Problem
Middle-clicking a placed furniture (a display entity with a barrier hitbox) in creative gives you
a `BARRIER` item instead of the Oraxen furniture item. This worked on older Minecraft versions
but broke after the MC 1.21.4 pick-block refactor.

## Root Cause
Since MC 1.21.4, Mojang split creative pick-block into two dedicated serverbound packets —
`ServerboundPickItemFromBlockPacket` and `ServerboundPickItemFromEntityPacket` — and the server
now resolves the picked item itself (`getCloneItemStack` + `tryPickItem`) **without firing any
Bukkit event**. Oraxen's existing middle-click handler in `FurnitureListener` listens on
`InventoryCreativeEvent`, which is never fired for picks anymore, so the player is handed the
furniture's hitbox block (a barrier) directly.

## Changes
- Added `PickItemHandler`: shared logic that checks whether the picked block/entity is an Oraxen
  furniture (`OraxenFurniture.getFurnitureMechanic`) and gives the player the real furniture
  item, replicating vanilla `tryPickItem` (selects an existing matching slot, otherwise adds the
  item in creative).
- Added `PickItemPacketListener` for both packet libraries Oraxen supports:
  - **ProtocolLib**: intercepts `PICK_ITEM_FROM_BLOCK` and `PICK_ITEM` (ProtocolLib maps the new
    entity pick onto the legacy `PICK_ITEM` id, so it is distinguished by packet class name).
  - **PacketEvents**: intercepts `PICK_ITEM_FROM_BLOCK` and `PICK_ITEM_FROM_ENTITY`.
- Wired both listeners through `PacketAdapter` / `PacketEventsAdapter` / `ProtocolLibAdapter` and
  registered them in `OraxenPlugin`.
- Both paths are Folia-safe: Bukkit block/entity access is deferred to the region thread, and the
  vanilla item is re-applied for non-furniture blocks on Folia (the packet must be cancelled
  up-front).
- Pre-1.21.4 servers keep working via the existing `InventoryCreativeEvent` handlers — the new
  listeners only match the new pick packets.

## Result
Middle-clicking an Oraxen furniture in creative now gives the actual furniture item (or selects
it from the hotbar if already present), on both ProtocolLib and PacketEvents setups. Verified
in-game on Paper.

### Question
@pxlarified  Would you prefer this PR and future ones are opened against `miziusLabs/OraxenPlugin/tree/1.220.0` instead? Happy to
move it over (and rebase onto that branch) if that's easier for you.****

### DIFF
--- a/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
+++ b/src/main/java/io/th0rgal/oraxen/OraxenPlugin.java
@@ -130,6 +130,7 @@ public class OraxenPlugin extends JavaPlugin {
             if (Settings.FORMAT_INVENTORY_TITLES.toBool())
                 packetAdapter.registerInventoryListener();
             packetAdapter.registerTitleListener();
+            packetAdapter.registerPickItemListener();
         });
 
         Bukkit.getPluginManager().registerEvents(new CustomArmorListener(), this);
--- a/src/main/java/io/th0rgal/oraxen/packets/PacketAdapter.java
+++ b/src/main/java/io/th0rgal/oraxen/packets/PacketAdapter.java
@@ -32,8 +32,10 @@ public interface PacketAdapter {
     void registerInventoryListener();
     void registerScoreboardListener();
     void registerTitleListener();
+    void registerPickItemListener();
     void removeInventoryListener();
     void removeTitleListener();
+    void removePickItemListener();
     void reregisterEfficencyMechanicListener(EfficiencyMechanicFactory efficiencyMechanicFactory);
 
     String getLatestMCVersion();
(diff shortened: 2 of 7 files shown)


pxlarified:
Sure, against miziusLabs/OraxenPlugin/tree/1.220.0 would be great.

OlivierChiasson:
PR re-opened at https://github.com/miziusLabs/OraxenPlugin/pull/4

pxlarified:
Thank you.

PR #1857, opened 2026-09-17
v1.219.1

pxlarified:
### Changes

- Added the vanilla `death_protection` item component (1.21.2+) which prevents death and supports configurable death effects.
  - Supports applying and removing effects, clearing effects, random teleportation, and playing sounds when death is prevented.
  - Preserves `death_protection` when existing Oraxen items are updated and exposes `death_effects` as an array in the generated schema.
- Added configurable unprotected resource-pack exports after generation and before obfuscation, while preserving normal client pack streams and upload routing.
  - Defaults to `./.output/unprotected.zip`, creates missing parent directories, and prevents exports inside the resource-pack source folder.
  - Writes through a temporary ZIP so failed exports do not replace the previous export and restores consumed streams for normal client-pack generation.
- Replaced several unnecessary NMS abstractions with shared Bukkit/Paper APIs, including entity ID allocation, food component handling, and jukebox stop behavior, and removed dead NMS listener hooks.
  - Furniture text-display packets now use PacketEvents directly.
  - Backpack and furniture-text virtual entity IDs now use Bukkit/Paper's entity ID allocation directly.

### Fixes

- Fixed written books failing to open on Minecraft/Purpur 26.2 with Adventure 5 due to the removed covariant `Book.Builder#build(): Book` method.
  - Uses the stable `Book.book(...)` factory, preserving compatibility with both Adventure 4.x and 5.x.
- Fixed furniture with an `itemname` losing its displayed name after being placed, broken, and picked up.
  - Clears the empty custom name used by placed furniture when rebuilding its drop, preserving configured custom names and `itemname` values.
- Fixed custom note-block models briefly rendering as vanilla note blocks when vertically adjacent custom or vanilla blocks are broken or placed.
  - Keeps vulnerable breaks server-authoritative while preserving configured mining timing, vanilla drops, tool damage, events, and protection listeners.
  - Replays authoritative target and custom-neighbor states in packet order after adjacent block changes and matches acknowledgements to the exact client prediction.
  - Preserves directional and waterlogged placement state, multi-block placement, item consumption, cancellation handling, Folia region ownership, and legacy fallbacks.
- Fixed furniture failing to break or interact correctly when persisted barrier offsets could not be deserialized because `BlockLocation` was not registered with Bukkit configuration serialization.
  - Registers `BlockLocation` for Bukkit configuration deserialization so persisted furniture barrier offsets can be read correctly when interacting with or attacking entity furniture.
- Prevented Oraxen items with `misc.allow_in_vanilla_recipes: false` from being used in automated crafter crafting.
  - Applies the same restriction as player crafting through `CrafterCraftEvent`.
  - Registers the crafter listener only when the server exposes the crafter API, preserving older-version compatibility.
  - Added regression coverage for restricted, allowed, ordinary, and absent misc mechanics.

### Maintenance

- Added regression coverage for preserving furniture item names across placement and drops on legacy and modern naming APIs.
- Added regression coverage for client-side note-block prediction, stacked custom block breaking, authoritative neighbor-state replay, and client-only mining-effect restoration.
- Added regression coverage for `BlockLocation` PDC serialization and deserialization.
- Added local beta/build scripts and a Paper/Folia test-server launcher, with the test server using port 25568.
- Updated the contribution list.
- Reduced startup and smithing listener method complexity.
- Bumped the project version to `1.219.1`.


Th0rgal:
Superseded by https://github.com/oraxen/oraxen/pull/1858 — same commits plus review fixes (custom noteblock placement acks, Folia actor packets, disabled unprotected-pack default, death_protection strip on update, prediction handler split, portable scripts).

